PR #9, opened 2021-04-01
Merge 30 or so existing tests from dauwhe/epub-tests

dlazin:
Here's a great baseline for further EPUB 3.3 tests.


iherman:
@dlazin you and @dauwhe are in control of this repo and, mainly in this initial setup phase, it sounds perfectly all right if you commit without an explicit approval...

I will merge it now.

PR #21, opened 2021-05-27
Add test for obfuscated resources (fonts)

dlazin:
Tests the following normative statement:

https://www.w3.org/TR/epub-rs-33/#confreq-ocf-fobfus
Reading Systems MUST support deobfuscation of resources as defined in Resource Obfuscation [EPUB-33].


iherman:
I did not check this, but is the font file obfuscated? Per the spec, the obfuscation must be done by the content creator...

iherman:
@dlazin I propose to close this PR in favour of #201, which certainly uses an obfuscated font.

PR #22, opened 2021-05-28
Add tests for package direction

dlazin:
Adds tests for the following normative statements:

https://w3c.github.io/epub-specs/epub33/rs/#sec-pkg-doc-base-dir
If the dir attribute is set and indicates a base direction of ltr or rtl, Reading Systems MUST override the bidi algorithm per the higher-level protocols defined in [BIDI], setting the paragraph embedding level to 0 if the base direction is ltr, or 1 if the base direction is rtl.

Otherwise the base direction is auto, in which case Reading Systems MUST determine the text's direction by applying the Unicode Bidi Algorithm, beginning with Rule P2 of [BIDI].

Note that both tests fail in Apple Books, Play Books (web), and Thorium, which are the only RSes I've checked them in.


iherman (commented):
Testing this is always a tough one, and I am not 100% sure that what is in the tests is correct (not being an expert either). Also, I do not find the example description "test passes if...." very clear to decide what should and should not be the expected behaviour.

I would use the following, somewhat shorter text (given to us by @r12a a while ago):

```
HTML و CSS: تصميم و إنشاء مواقع الويب
```

Technically, the problem with this text is that the first character in memory is the letter "H", and, if the value of `dir` is `auto`, then the Unicode algorithm will determine that the text direction is `ltr`. As a result, this is what is the expected behaviour (and I use images for the expected results to make them clear, probably this is what we should do in the test as well):

Screen Shot 2021-05-31 at 12 06 31

**_This rendering is wrong_**, but that is the expected behaviour for `auto`! 

By adding `dir=rtl`, the same display will be:

Screen Shot 2021-05-31 at 12 07 18

which **is** the right behaviour: the `dir=rtl` attribute instructs the renderer to ignore automatic procedure.

Using these strings might be better for the tests because they are shorter and make the point clearer.

Actually, for the first test (testing auto) it may be even clearer if we just used an Arabic text without further additional complication. The point is that the full text should be displayed correctly without any further intervention, because the first character is arabic and, therefore, the text should be then right-to-left.

(We may want to make it a bit more realistic and choose a real book title with a real creator...)


(@r12a might chime in).

iherman:
B.t.w., it may be better to use Hebrew texts, if we can find one, than Arabic. The cursive nature of Arabic may make it more difficult for non-Arabic/non-Hebrew speakers to discern the characters and their positioning... Maybe @TzviyaSiegman can provide us with good examples...

mattgarrish:
While I don't want to have an excessive number of tests, what if a reading system has only implemented partial support? (e.g., for title and author, but not for descriptions) Is that a blanket fail or does it pass because we've got part of the behaviour exhibited?

I'm not arguing for atomizing the `dir` tests for each piece of metadata, but some instruction that a partial pass counts as an implementation might help, assuming we'll accept that.

From past experience with these kinds of tests, prose descriptions aren't always helpful, especially for non-native speakers. It's generally helpful to also include a screen grab of the expected rendering so that the result can also be visually verified.

Also, I don't see an ltr example looking at the source, but we'll be globally testing reading systems. It would be good to check that reading systems don't assume rtl in all cases. There should also be a check for an explicit `dir="auto"` as that value is new in 3.3.

I'm not sure what depth of coverage you're going for in this PR, so apologies if I'm listing things you're going to get to, but what about also testing overrides on the individual metadata elements? We should have some tests for overriding from ltr to rtl, ltr to rtl, etc.

iherman:
I think the six directional tests that have been added to #27 makes this test unnecessary. I would propose to close this one, overtaken by events...

dlazin:
> I think the six directional tests that have been added to #27 makes this test unnecessary. I would propose to close this one, overtaken by events...

Agreed, except now it's https://github.com/w3c/epub-tests/pull/31.

PR #23, opened 2021-08-26
Initial testing instructions

dlazin:
Write initial testing instructions, incorporating Ivan's docs from https://github.com/iherman/epub-testing/blob/testing-action-via-PR/docs/formats.md.

Staged:
https://github.com/w3c/epub-tests/tree/instructions


wareid (approved):
I have some things I'd like to add to this, like instructions for sideloading on kobo apps and devices that might be useful, as well as an EPUB zipping terminal command that works perfect on modern MacOS systems, since eCanCrusher doesn't. Will add as a commit.

iherman:
What is not 100% clear is the organization of the test folder itself. The current repository seems to have:

- `tests/EPUBs`, which is a flat collection of individual tests in EPUB
- `tests/reading-systems`, which is a hierarchy of tests folders

What is the relationship between the two? Is the first a 'flattened' version of the directories in the second? What does the hierarchy in the second correspond to?

As a test developer, I may prefer the `tests/reading-systems` replaced by something like `tests/directories` or something similar, which has a 1-1 correspondence with the EPUB-s themselves. After all, tests may evolve, and always start with the EPUB and unzip it before doing a modification may be a drag. Also, for a casual reader, say, on Github, looking at the directory directly is easier than into the epub.

---

(B.t.w., my current reporting mechanism also relies on a flat collection of test directories rather than epubs. I can, of course add a layer to my script to unzip the epub file to extract the metadata, but that is an extra step that I would have to add. To be clear: no big deal, I can do that, but only if we decide it is unnecessary to keep the directories in the repositiory.

iherman:
The weak point in my metadata scheme is `dc:coverage`. I believe we should maintain a list of 'coverage' ourselves, sometimes on the repository, otherwise it may become the Wild West...

iherman:
The issue was discussed in a [meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-09-10-epub#section1) on 2021-09-10

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

### 1. Initial testing instructions (pr epub-tests#23)

_See github pull request [epub-tests#23](https://github.com/w3c/epub-tests/pull/23)._




**Dan Lazin:** I've prepared a document on testing instructions  
… then we have to put metadata so we can create automatic reports from the metadata  
… so we can invite non-testers to start testing  
… we are looking for volunteers for creating tests for FXLs and MOs EPUBs  

**Ivan Herman:** I have a system for automatic reporting but only in a test repository, I will add it to the tests' repository when it is final in its structure  
… this system will create reports automatically for every new implementer  

**Dave Cramer:** can you describe how the reports will work?  

**Ivan Herman:** every implementer should submit a JSON file (super simple structure) giving a true or false if the test worked or not  
… then the system will get all the results and create the report  
… it's really similar to the one used for audiobooks  

**Brady Duga:** what it means that a reading systems pass a test?  
… I have three reading systems, what happen if one passes and the others not?  

**Ivan Herman:** It's up to the WG or to the implementer to decide if it is a unique reading system or that are three different reading systems  
… the goal of the testing is not a beauty contest, the goal is to check that the features in the spec are consistent and can be implemented  

**Dave Cramer:** from my point of view, the google play books is one system (with three implementations)  

**Brady Duga:** we have three different teams, the clients are different, but the backend and the pipeline is common for all three  
… so maybe yes, it's only one system  
… but what does happen if one implementation pass and the other two doesn't?  

**George Kerscher:** for accessibility we differentiate the OSs, since AT + OS + reading system may have really different results  

**Wendy Reid:** I think that in the instructions we should say that depending on how you have implemented you systems, it may be worth to test the different apps on different OSs  

> *Brady Duga:* +1 to wendyreid

**Tzviya Siegman:** I think the epubtest.org is super important to understand how reading systems work on different OSs  
… it's similar to ARIA: ARIA is not tested on the browser, but on the Accessibility API exposed by the browser engine  
… I don't know if it is possible with EPUBs  

> *Tzviya Siegman:* [https://w3c.github.io/test-results/wai-aria/all.html](https://w3c.github.io/test-results/wai-aria/all.html) ARIA Tests

**Brady Duga:** I don't know if it's meaningful: some work is managed by the backend, other is managed by the frontend  

**Ivan Herman:** I propose to report results in two formats: one for the W3C (unique implementations), the other for the community (separating the OSs). This will require additional metadata in the implementation report, but it should be doable  

> *Brady Duga:* +1 to ivan

</details>

dlazin:
> The weak point in my metadata scheme is `dc:coverage`. I believe we should maintain a list of 'coverage' ourselves, somewhere on the repository, otherwise it may become the Wild West...

Agreed, but I think we can add that as we proceed with the reorganization / metadata tagging. For now we'd just be guessing, and I think the appropriate structure will become clear once we're doing the real work. I'll make sure to add this in my reorg PR(s).

> What is not 100% clear is the organization of the test folder itself. The current repository seems to have:
> 
> * `tests/EPUBs`, which is a flat collection of individual tests in EPUB
> * `tests/reading-systems`, which is a hierarchy of tests folders
> 
> What is the relationship between the two? Is the first a 'flattened' version of the directories in the second? What does the hierarchy in the second correspond to?

I'm cutting the hierarchy (it is documented prior to this PR, and deleted in this). I have unsent local changes that flatten the directories in the second, so that you get this:

* `tests/cmt-gif.epub`
* `tests/cmt-gif/` (uncompressed source)

Whether that will work in the long term, or be too big to navigate, remains to be seen, but I propose we start like that. Will that work with your script?

> As a test developer, I may prefer the `tests/reading-systems` replaced by something like `tests/directories` or something similar, which has a 1-1 correspondence with the EPUB-s themselves. After all, tests may evolve, and always start with the EPUB and unzip it before doing a modification may be a drag. Also, for a casual reader, say, on Github, looking at the directory directly is easier than into the epub.

Yes, agreed, that's what my reorg does. I was trying to finish it before sending, but it might be more productive just to send the partial reorg that I've completed so far.

> (B.t.w., my current reporting mechanism also relies on a flat collection of test directories rather than epubs. I can, of course add a layer to my script to unzip the epub file to extract the metadata, but that is an extra step that I would have to add. To be clear: no big deal, I can do that, but only if we decide it is unnecessary to keep the directories in the repository.

Agreed, unnecessary; only open question is whether you need this:

* `tests/cmt-gif.epub`
* `tests/cmt-gif/` (uncompressed source)

Or this:

* `tests/cmt-gif.epub`
* `tests/directories/cmt-gif/` (uncompressed source)

I prefer the former. It's clear and works easily with eCanCrusher.

dlazin:
> I'm cutting the hierarchy (it is documented prior to this PR, and deleted in this). I have unsent local changes that flatten the directories in the second, so that you get this:
> 
> * `tests/cmt-gif.epub`
> * `tests/cmt-gif/` (uncompressed source)

I have now sent these proposed changes as https://github.com/w3c/epub-tests/pull/24, and incorporated your latest `variant` option into this doc.

iherman:
> > The weak point in my metadata scheme is `dc:coverage`. I believe we should maintain a list of 'coverage' ourselves, somewhere on the repository, otherwise it may become the Wild West...
> 
> Agreed, but I think we can add that as we proceed with the reorganization / metadata tagging. For now we'd just be guessing, and I think the appropriate structure will become clear once we're doing the real work. I'll make sure to add this in my reorg PR(s).

+1

> 
> > What is not 100% clear is the organization of the test folder itself. The current repository seems to have:
> > 
> > * `tests/EPUBs`, which is a flat collection of individual tests in EPUB
> > * `tests/reading-systems`, which is a hierarchy of tests folders
> > 
> > What is the relationship between the two? Is the first a 'flattened' version of the directories in the second? What does the hierarchy in the second correspond to?
> 
> I'm cutting the hierarchy (it is documented prior to this PR, and deleted in this). I have unsent local changes that flatten the directories in the second, so that you get this:
> 
> * `tests/cmt-gif.epub`
> * `tests/cmt-gif/` (uncompressed source)
> 
> Whether that will work in the long term, or be too big to navigate, remains to be seen, but I propose we start like that. Will that work with your script?

I can get that working. Once I close the 'variant' PR I can do that. It does not seem to be difficult.

>> 
> Agreed, unnecessary; only open question is whether you need this:
> 
> * `tests/cmt-gif.epub`
> * `tests/cmt-gif/` (uncompressed source)
> 
> Or this:
> 
> * `tests/cmt-gif.epub`
> * `tests/directories/cmt-gif/` (uncompressed source)
> 
> I prefer the former. It's clear and works easily with eCanCrusher.

I can do the former.

iherman (approved):

PR #27, opened 2021-09-21
Incorporated the reporting mechanism

iherman:
I have carried over the reporting mechanism

- The automatic reporting (ie, the github actions) have not yet been carried over; this really makes sense only when we have a meaningful report to generate. At this moment most of the tests have identical ID-s and that creates meaningless tables.
- There are two test implementations in the relevant folder, one with variants and one not. It would be good to have 'real' implementations put there soon, even if incomplete. The obvious candidates are google play, kobo, and thorium if we are allowed to use them by running the tests through them ourselves.
- While I was at it, I have also added the 6 tests I created a while ago (and I used in my own testings) for the direction attributes in the package metadata

Cc @dauwhe @wareid @bduga @dlazin


results/formats.md
```diff
@@ -0,0 +1,86 @@
+# Test and test report formats
```
dlazin, line 1: Remove this file, please; this is all covered in readme.md.

(Have you added anything new here in the last couple days? If so, please add to readme.md instead.)

src/lib/types.ts
```diff
@@ -0,0 +1,138 @@
...
+    export const TEST_RESULTS_DIR: string = 'reports';
+
+    /** Location for the final report */
+    export const DOCS_DIR: string = 'results';
+
+    /** Location for the implementation report templates */
+    export const TEST_RESULTS_TEMPLATE: string = `${TEST_RESULTS_DIR}/xx-template.json`;
+
+    /** Base URL for the tests in the repository */
+    export const TEST_URL_BASE: string = 'https://github.com/iherman/epub-testing/tree/main/tests';
```
dlazin, line 19: Does this need to be updated?

tests/confreq-mathml-rs-behavior/OPS/package.opf
```diff
@@ -4,7 +4,7 @@
     <dc:title>Title</dc:title>
     <dc:creator>Creator</dc:creator>
     <dc:date>2021-01-01</dc:date>
-    <dc:identifier id="pub-id">NOID</dc:identifier>
+    <dc:identifier id="pub-id">confreq-mathml-rs-behavior</dc:identifier>
```
dlazin, line 7: Why is this change in this PR? Seems like it should be in #29.

tests/package-title-dir-rtl-001/OPS/content_001.xhtml
```diff
@@ -1,13 +1,18 @@
 <html xmlns="http://www.w3.org/1999/xhtml" xmlns:epub="http://www.idpf.org/2007/ops" xml:lang="en">
```
dlazin, line 1: Optional, but I'd recommend putting this new test in a separate PR — unrelated. I won't review this test carefully in this PR.

tests/package-title-dir-rtl-001/OPS/package.opf
```diff
@@ -1,18 +1,23 @@
...
-</metadata>
-<manifest>
-  <item id="content_001" href="content_001.xhtml" media-type="application/xhtml+xml" />
-  <item id="nav" properties="nav" href="nav.xhtml" media-type="application/xhtml+xml" />
-</manifest>
-<spine>
-  <itemref idref="content_001" />
-</spine>
+    <metadata xmlns:dc="http://purl.org/dc/elements/1.1/">
+        <dc:title dir="rtl" xml:lang="he">CSS: הרפתקה חדשה!</dc:title>
```
dlazin, line 4: My existing tests (cmt-* and xx-*) alphabetize all the metadata; consider matching.

tests/package-title-dir-rtl-001/OPS/package.opf
```diff
@@ -1,18 +1,23 @@
...
+        <dc:date>2021-06-24</dc:date>
+        <dc:creator>Ivan Herman</dc:creator>
+        <dc:identifier id="pub-id">package-title-dir-rtl-001</dc:identifier>
+        <meta property="dcterms:modified">2021-06-24T00:00:00Z</meta>
+        <dc:language>en</dc:language>
+        <dc:coverage>Internationalization Features</dc:coverage>
```
dlazin, line 10: As in #29, I recommend calling this just Internationalization.

dlazin (changes requested):

results/formats.md
iherman: Yep will do in the next commit

src/lib/types.ts
iherman: Ouch, good catch!

tests/confreq-mathml-rs-behavior/OPS/package.opf
iherman: That is a side-effect of my own testing of the script. With the merge of #29 this is now done.

tests/package-title-dir-rtl-001/OPS/content_001.xhtml
iherman: You are right. Let us play by the rules and add these via a separate PR. Will do.

tests/package-title-dir-rtl-001/OPS/package.opf
iherman: Moot by moving these into a separate PR (but will do there).

tests/package-title-dir-rtl-001/OPS/package.opf
iherman: Moot by moving these into a separate PR (but will do there).

PR #29, opened 2021-09-22
Adding valid ID values to the tests

iherman:
This is a minimal change for all tests (when applicable) that adds the required ID values to all tests. This makes the generated test reports a bit more "real".

Only the ID values have been added (unless it was set), except for the progression direction test that has been added to the "internationalization features" category.

Cc @dauwhe @dlazin @wareid


iherman:
Note that this is a PR against PR #27, ie, the one adding the reporting features!

dauwhe (approved):

tests/package-creator-dir-rtl-001/OPS/package.opf
```diff
@@ -4,8 +4,9 @@
   <dc:title>Package Creator Direction RTL 001</dc:title>
   <dc:date>2021-02-03</dc:date>
   <dc:creator dir="rtl">غ ظ ض ذ خ ث ت ش ر ق ص ف ع س ن م ل ك ي ط ح ز و ه د ج ب أ</dc:creator>
-  <dc:identifier id="pub-id">NOID</dc:identifier>
+  <dc:identifier id="pub-id">package-creator-dir-rtl-001</dc:identifier>
   <meta property="dcterms:modified">2021-02-03T00:00:00Z</meta>
+  <dc:coverage>Internationalization Features</dc:coverage>
```
dlazin, line 9: Can we call this just Internationalization instead? "Features" is at least implicit, and sometimes maybe incorrect.

Please also update the list of coverage values here:
https://github.com/w3c/epub-tests/blob/main/README.md#metadata

dlazin (approved):

tests/package-creator-dir-rtl-001/OPS/package.opf
iherman: Yep, I agree. I will commit the changes and then merge

PR #31, opened 2021-09-22
Added text base direction tests

iherman:
Tests on the bidirectionality control in the package metadata


wareid (approved):

tests/package-title-dir-rtl-001/OPS/content_001.xhtml
```diff
@@ -0,0 +1,18 @@
+<html xmlns="http://www.w3.org/1999/xhtml" xmlns:epub="http://www.idpf.org/2007/ops" xml:lang="en">
```
dlazin, line 1: Nit: remove blank lines and adjust indentation to match https://github.com/w3c/epub-tests/blob/main/tests/xx-epub-template/OPS/content_001.xhtml?

I changed it a few PRs ago.

tests/package-title-dir-rtl-001/OPS/package.opf
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<package xmlns="http://www.idpf.org/2007/opf" xmlns:epub="http://www.idpf.org/2007/ops" version="3.0" xml:lang="en" unique-identifier="pub-id">
+    <metadata xmlns:dc="http://purl.org/dc/elements/1.1/">
```
dlazin, line 3: Super-nit: this appears to use 4-space indentation (unless this is a GitHub review artifact), whereas https://github.com/w3c/epub-tests/blob/main/tests/xx-epub-template/OPS/package.opf does 2 spaces.

Obviously, this doesn't actually matter, but if it's all the same, I prefer consistency.

If your editor is generating the 4 spaces and it's a pain to change, then ignore, but if it's arbitrary then let's be consistent if possible.

tests/package-title-dir-rtl-001/OPS/package.opf
```diff
@@ -0,0 +1,23 @@
...
+        <dc:coverage>Internationalization</dc:coverage>
+        <dc:creator>Ivan Herman</dc:creator>
+        <dc:date>2021-09-22</dc:date>
```
dlazin, line 6: Opinion: should we have real dates here? I kinda like them, but we do not have real dates in other tests. Everything else is 2021-01-01.

tests/package-title-dir-rtl-001/OPS/package.opf
```diff
@@ -0,0 +1,23 @@
...
+        <dc:description>The 'dc:title' element contains a text whose proper rendering requires bidi control. The element's 'dir' attribute is set to 'rtl'; the title should display correctly.</dc:description>
```
dlazin, line 7: Nit: instead of "correctly," say ... "from right to left," I guess?

tests/package-title-dir-rtl-001/OPS/package.opf
```diff
@@ -0,0 +1,23 @@
...
+        <dc:identifier id="pub-id">package-title-dir-rtl-001</dc:identifier>
+        <dc:language>en</dc:language>
+        <dc:title dir="rtl" xml:lang="he">CSS: הרפתקה חדשה!</dc:title>
+        <meta property="dcterms:alternative">Correct 'dir' on the title</meta>
```
dlazin, line 11: Since there should be several dir tests (right?), should this say "Correct rendering of 'dir="rtl"' on the title"? I think you will need an ltr case later, an auto value, an implicit auto, etc.

tests/package-title-dir-rtl-001/META-INF/container.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
dlazin, line 1: I think this directory name does not match the proposed naming in https://github.com/w3c/epub-tests#naming (which is open for discussion).

Based on that, I think the directory name should be confreq-rs-pkg-dir_rtl (and similar for the other tests).

If you agree, please update the test ID to match, of course.

tests/package-title-dir-rtl-001/OPS/content_001.xhtml
```diff
@@ -0,0 +1,18 @@
...
+
+<head>
+    <title>Package Title Direction RTL 001</title>
```
dlazin, line 4: Nit: I'm not really setting the XHTML <title> element (do RSes ever show it?), but if we're going to, I think it should be similar to the test ID, rather than 001. That is, something something dir="rtl".

dlazin (commented):
Sorry for all the comments, but I want to get on the same page early :)

tests/package-title-dir-rtl-001/OPS/package.opf
iherman: Dates are currently not used, that is true, but I myself would prefer to use real dates when possible. It may help us, later, to trace back when the tests have been done...

tests/package-title-dir-rtl-001/OPS/package.opf
iherman: +1. Similar changes on the titles of all the others, too.

tests/package-title-dir-rtl-001/META-INF/container.xml
iherman: I have changed to `confreq-rs-pkg-dir_rtl-001` (and -002, 003, etc.). But it is not really clear cut. Some of the tests are related to the interplay between `dir=rtl` and `dir=ltr` or `dir=auto`, ie, one could also consider to use `confreq-rs-pkg-dir-auto-001` (there is a separate `confreq-rs-pkg-dir-auto` anchor in the spec). But I presume we will have such border cases elsewhere, so the naming should be a bit flexible and, at the end of the day, the best judgement should prevail...

tests/package-title-dir-rtl-001/OPS/content_001.xhtml
iherman: Ah, I see that you have not set the title in cmt tests, just left it 'Title'... However, the fact is that test providers (at least I) also look at the content in a traditional browser to check whether it is all right, and therefore the title may be helpful (you are right that RS-s ignore it).

Using the title as appear in the package is a good idea. That makes it clean.

tests/package-title-dir-rtl-001/OPS/content_001.xhtml
iherman: Done it in https://github.com/w3c/epub-tests/pull/31/commits/dc9f0d396bd4d814812b3d707225bef4f0f59357

tests/package-title-dir-rtl-001/OPS/package.opf
iherman: Done it in https://github.com/w3c/epub-tests/pull/31/commits/dc9f0d396bd4d814812b3d707225bef4f0f59357

tests/package-title-dir-rtl-001/OPS/package.opf
iherman: Done it in https://github.com/w3c/epub-tests/pull/31/commits/dc9f0d396bd4d814812b3d707225bef4f0f59357

tests/package-title-dir-rtl-001/OPS/package.opf
iherman: Done it in https://github.com/w3c/epub-tests/pull/31/commits/dc9f0d396bd4d814812b3d707225bef4f0f59357

tests/package-title-dir-rtl-001/OPS/package.opf
iherman: Done it in https://github.com/w3c/epub-tests/pull/31/commits/dc9f0d396bd4d814812b3d707225bef4f0f59357

tests/package-title-dir-rtl-001/META-INF/container.xml
iherman: Done it in https://github.com/w3c/epub-tests/pull/31/commits/dc9f0d396bd4d814812b3d707225bef4f0f59357

tests/package-title-dir-rtl-001/OPS/content_001.xhtml
iherman: Done it in https://github.com/w3c/epub-tests/pull/31/commits/dc9f0d396bd4d814812b3d707225bef4f0f59357

tests/cmt-jpeg/OPS/content_001.xhtml
```diff
@@ -1,6 +1,6 @@
 <html xmlns="http://www.w3.org/1999/xhtml" xmlns:epub="http://www.idpf.org/2007/ops" xml:lang="en">
   <head>
-    <title>Title</title>
+    <title>Correct 'dir=rtl' on the title</title>
```
dlazin, line 3: This is a cmt-jpeg test. Mistake?

tests/package-title-dir-rtl-001/META-INF/container.xml
dlazin: Sounds good to me.

dlazin (approved):

PR #32, opened 2021-09-22
Adding the test creators to the report.

iherman:
The additions are [in the test description file](https://raw.githack.com/w3c/epub-tests/Adding-test-creators/results/tests.html).


results/fragments/creators.html
```diff
@@ -0,0 +1,4 @@
+<ul class="creator_list">
```
dlazin, line 1: Super-nit: throughout, my instinct would be to call this test_authors rather than creators, but I don't think it's user-visible, so up to you.

dlazin (approved):

results/fragments/creators.html
iherman: I was motivated by the fact that the DC term used is 'creator' and not authors. That is why I systematically used the term 'creator' in the code for different variables. I would prefer to leave it that way...

PR #33, opened 2021-09-22
Add test for fixed layout properties 

wareid:
Adding my first test for fixed layout properties. 

(This is also a practice run for the test I'll write on Friday).

PR #37, opened 2021-09-24
Adding test for fixed layout properties 

wareid:
Adding test for fxl-properties-002. Please review.


dlazin (commented):
Is this the same as https://github.com/w3c/epub-tests/pull/33, just from today's demo? If so, which would you like me to review? If not the same, can you explain what's different here?

PR #41, opened 2021-09-29
Respec is run as part of the action

iherman:
The respec command-line processor is installed as part of the action and is run to generate the final reports. This means that, for the end user, the reports are displayed without drag and with just the standard cache. The price is that the run of the action is a bit longer, but that is not really a problem.

As part of the changes I have also added new triggers to the action: it is also run if the core script changes or if the respec sources for the reports are changed.

Note that this required a little reorganization: there is a `drafts` directory under `results`, which contains the respec sources and the `drafts/fragment` files are generated by the script. The html files in the `results` directory are now the generated ones.


dlazin (approved):
Excellent.

iherman:
The merge conflicts were too complicated to handle it here, so I made the necessary actions on my machine and committed the 'main' branch directly. This replaces the merge using this tool...

Closing without any further action here.

Issue #68, opened 2021-10-30
Should we add a 'rights' entry to the metadata?


iherman:
This issue came up during the [2nd vF2F day](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-10-29-epub): we may want to add a metadata item into the `package.opf` file, namely:

```xml
<dc:rights>https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document</dc:rights>
```

The reason is that some reading systems will have to put a copy of the tests into another public space for testing, and the questions on rights may arise.

(Caveat: we would have to add this line to all opf files. Wonder whether this could be automatized; I am willing to look into this in case we otherwise agree.)

iherman:
@wseltzer your reaction on this would be important!

iherman:
If we want to be complete, we can also add a 'rightsHolder' term, too, pointing at W3C.

iherman:
> If we want to be complete, we can also add a 'rightsHolder' term, too, pointing at W3C.

While this would be correct in terms of DC data, epubcheck complains; it is not an allowed metadata item.

I have a code on my machine which seems to be ready to go to regenerate all the tests with the rights metadata added. Should I?

@dauwhe @wareid @dlazin

wseltzer:
@iherman can you share a pointer to a bit more description? What is it we'd be asserting with that statement? 
It seems uncontroversial if it's just linking the terms of the license.

mattgarrish:
> While this would be correct in terms of DC data, epubcheck complains; it is not an allowed metadata item.

It's not an element in the DC set, if that's what you mean, but there's no reason you can't specify these as rel values on a `link`:

```
<link rel="dcterms:rights" href="https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document"/>
<link rel="dcterms:rightsHolder" href="..."/>
```

`meta` tags always strike me as ugly for URLs, but you could do it that way, too:

```
<dc:rights>https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document</dc:rights>
<meta property="dcterms:rightsHolder">...</meta>
```

dlazin:
I am cool with putting such metadata in every package. But are you talking about writing it programmatically as some sort of post-processing, or just a bulk GitHub edit? Bulk edit sounds good to me. I am uncomfortable with writing things into the EPUBs that are not visible in the uncompiled source, though.

iherman:
[...]

PR #69, opened 2021-11-04
Adding new metadata globally

iherman:
This is a PR for fixing #68.

@dlazin and others: I do not ask you to review all the test metadata; by looking at one or two you've seen them all (the same action is done on all OPF files). That being said, it would be welcome if you just have a quick look at the typescript code changes; you have a great eye for picking out problems and inconsistencies...

The script has also regenerated the epub file themselves.

Added bonus: I deliberately did the code in a way to be able to do similar actions, in the future, for other things as well. In particular, I can now generate the epub files from the directory from inside the typescript code:-)

Fis #68


src/add_metadata.ts
```diff
@@ -0,0 +1,72 @@
...
+const fs = fs_old_school.promises;
+
+import {get_list_dir, isDirectory} from './lib/data';
+import { Constants } from './lib/types';
+import { create_epub } from './lib/epub';
+
+/**
+ * The extra metadata items to be added to the package metadata
+ */
+const new_metadata: string[] = [
```
dlazin, line 23: Nit: ideally I think this would be a flag that you pass, since these elements will have to be removed the next time we want to use this script. But that's a change we can make next time :)

dlazin (approved):
Excellent! Thanks.

src/add_metadata.ts
iherman: While I was at it, I made the change in the code. If ever we have to reuse this code, other changes may be necessary, but better get this out of the way:-)

wareid (approved):

dauwhe (approved):

PR #73, opened 2021-11-07
Rename all tests as discussed in testing meeting

dlazin:
This is not a pure rename; it also makes a few small corrections/clarifications for things I discovered along the way. But close enough.


iherman (approved):
Let us hope this is the last test rename step... :-)

iherman (commented):
There is a problem with renaming.

The (new) `contributing.md` file says, for renaming:

> Test names should start with a three-letter abbreviation that corresponds to the value of the dc:coverage element below (for example, cnt for Content Documents, pkg for Package Documents, etc.), followed by a short hyphenated identifier that makes clear which requirement is under test. 

But that does not seem to be followed in, say, `pub-cmt-mp3` which has "Core Media Types" as the coverage value. The name should then be `cmt-mp3`, shouldn't it?

Also, 'three letter abbreviation' is a bit arbitrary. It seems that `pub` stands for tests on Publication Resources, ie, the abbreviation is the first three letters of the first word, whereas for the Core Media Types it is `cmt`?

dlazin:
> There is a problem with renaming.
> 
> The (new) `contributing.md` file says, for renaming:
> 
> > Test names should start with a three-letter abbreviation that corresponds to the value of the dc:coverage element below (for example, cnt for Content Documents, pkg for Package Documents, etc.), followed by a short hyphenated identifier that makes clear which requirement is under test.
> 
> But that does not seem to be followed in, say, `pub-cmt-mp3` which has "Core Media Types" as the coverage value. The name should then be `cmt-mp3`, shouldn't it?
> 
> Also, 'three letter abbreviation' is a bit arbitrary. It seems that `pub` stands for tests on Publication Resources, ie, the abbreviation is the first three letters of the first word, whereas for the Core Media Types it is `cmt`?

Yes, known issue :) What do you think we should do about it? In practice, I think it's fine for us to be semi-arbitrary and just pattern-match off existing tests, but that's hard to put in documentation. (Well, weird. We *can* just tell people to follow what they see.)

There are a couple inconsistencies here:
1. The existing Internationalization section doesn't match the structure of the specs. So `coverage` is Internationalization but I have the tests named `pkg-*` etc to correspond to the section of the spec.
2. Same thing for Core Media Types — it's a section in the test doc but not the spec. I have renamed those to `pub-cmt-*` (prior to your comment, unless I missed something and that's what you're seeing). Note that Content Documents are `cnt-*`.

I'm going to commit this to avoid merge problems, but I am happy to tweak the documentation and maybe rename a few more tests (should we get rid of the Internationalization and Core Media Types sections?) once we have consensus.

iherman:
> > The (new) `contributing.md` file says, for renaming:
> > > Test names should start with a three-letter abbreviation that corresponds to the value of the dc:coverage element below (for example, cnt for Content Documents, pkg for Package Documents, etc.), followed by a short hyphenated identifier that makes clear which requirement is under test.

> > Also, 'three letter abbreviation' is a bit arbitrary. It seems that `pub` stands for tests on Publication Resources, ie, the abbreviation is the first three letters of the first word, whereas for the Core Media Types it is `cmt`?
> 
> Yes, known issue :) What do you think we should do about it? In practice, I think it's fine for us to be semi-arbitrary and just pattern-match off existing tests, but that's hard to put in documentation. (Well, weird. We _can_ just tell people to follow what they see.)
> 
> There are a couple inconsistencies here:
> 
> 1. The existing Internationalization section doesn't match the structure of the specs. So `coverage` is Internationalization but I have the tests named `pkg-*` etc to correspond to the section of the spec.
> 2. Same thing for Core Media Types — it's a section in the test doc but not the spec. I have renamed those to `pub-cmt-*` (prior to your comment, unless I missed something and that's what you're seeing). Note that Content Documents are `cnt-*`.
> 

I am fine with some inconsistency; after all, the names are more or less opaque at the end, they are merely identifiers. The only thing I am worried about is what name a lambda test writer should choose, ie, whether this information is clear. Maybe simply maintaining a list of the existing abbreviations related to the respective section in the spec is good enough.

> I'm going to commit this to avoid merge problems, 

+1

> but I am happy to tweak the documentation and maybe rename a few more tests (should we get rid of the Internationalization and Core Media Types sections?) once we have consensus.

I think having significant subsections makes sense. I am not sure about the core media type one, but I think keeping the I18N one is more important. I guess it will be a judgement call every time...

PR #76, opened 2021-11-09
Added the references to the new metadata in the documentation

iherman:
I forgot to do that when generating the new metadata items

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -158,6 +158,11 @@ The package document for each test must contain the following metadata, which is
   report. This should be used if the subject of the test is the value of `dc:title` itself (e.g., testing the base direction
   of the `title` element).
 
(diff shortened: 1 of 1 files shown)


dlazin (approved):
This is fine if you want to do it, but I deliberately did not do this because your intro to the code block says:
`In this example, only the relevant metadata items are shown:`

My reading of that was that since the rights and rightsHolder elements are standard across all tests, they do not need to be documented here, but up to your judgment!

iherman:
I see two reasons why it would be good to have this:

- Some tests may come from the "outside" and incorporated into the system and not start from templates. Eg., if we carry over a11y tests that we discussed on the F2F
- There may be a remote possibility that someone comes with a test but, for whatever reasons, asks to have a different right statement...

PR #79, opened 2021-11-12
Add test for relative URLs

### LINKED ISSUE #59: Test package documents with arbitrary filenames
https://w3c.github.io/epub-tests/#container-default-rendition_arbitrary tests package documents in an arbitrary directory (`FOO/BAR/package.opf`), but we don't have any tests of packages with arbitrary *filenames* (`foo.opf` and maybe `foo.bar` or just `foo`).

That should be added as an additional test case for https://w3c.github.io/epub-specs/epub33/rs/#container-default-rendition ("A Reading System MUST, by default, use the Package Document referenced from first rootfile element to render the EPUB Publication.")

### LINKED ISSUE #78: Test parsing relative URLs
I will test this:
https://w3c.github.io/epub-specs/epub33/rs/#sec-pkg-doc-relative-urls

dlazin:
This tests the following normative statement:
> To parse relative-URL-with-fragment strings [URL] in the Package Document, Reading Systems MUST use the URL of the Package Document as the base URL [URL].

> When an EPUB Publication is zipped, the base URL of the Package Document is obtained from the URL of the EPUB Container together with a fragment identifier that specifies the path to Package Document (relative to the Root Directory). This specification does not require a specific URL scheme for referencing the path to the Package Document within the EPUB Container.

https://w3c.github.io/epub-specs/epub33/rs/#sec-pkg-doc-relative-urls

Resolves #78, resolves #59!


iherman (commented):
The test itself looks o.k. to me. However, we should hold off this tests until https://github.com/w3c/epub-specs/pull/1898 is finalized and merged, because all the references may change. This comment is particularly true for https://github.com/w3c/epub-specs/pull/1908 which would probably become invalid as such

wareid (approved):

dlazin:
Do you think this test is still good? I haven't read w3c/epub-specs#1898 carefully. I will go resolve the conflicts in https://github.com/w3c/epub-specs/pull/1908 now.

tests/pkg-relative-url/foo/BAR/baz.opf
```diff
@@ -0,0 +1,26 @@
...
+    	The manifest refers to an XHTML file in an arbitrary subfolder that is relative to the package's own arbitrary folder.
+      The reading system must be able to find the content.
+    </dc:description>
+    <dc:identifier id="pub-id">pkg-relative-url</dc:identifier>
+    <dc:language>en</dc:language>
+    <dc:title>Content in relative subdirectory</dc:title>
+    <link rel="dcterms:rights" href="https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document"/>
+    <link rel="dcterms:rightsHolder" href="https://www.w3.org"/>
+    <meta property="dcterms:isReferencedBy">https://www.w3.org/TR/epub-rs-33/#confreq-rs-pkg-relative-url</meta>
+    <meta property="dcterms:modified">2021-21-12T00:00:00Z</meta>
```
iherman, line 17: This reference is now wrong.

For the RS document, it is not 100% clear where to put the reference, but I suspect it should be bound to the first paragraph of [§3.4 Manifest](https://www.w3.org/TR/epub-rs-33/#sec-pkg-doc-manifest). (We may need an extra id to the paragraph, though).

The "real" target is probably https://www.w3.org/TR/epub-33/#sec-container-iri, though, more exactly its first paragraph.

I would probably add both references.

dlazin: Ah, thanks for catching that — done. The converse references are in https://github.com/w3c/epub-specs/pull/1908.

iherman (approved):

PR #86, opened 2021-11-18
Update test IDs in implementation reports

dlazin:
This is approximately the same as you had the data, but since it's mostly fake, I only did best-effort conversion. For T, I did a slightly better job of mapping, but nonetheless I simply deleted (so N/A) some tests whose mappings I could not remember, since this is still just an example and will need to be retested. Let me know if you'd like me to do a perfect job with T.


iherman (approved):
That is o.k. At some point, when we get to an equilibrium point and I have the time for it, I may run through T more systematically although, really, they shoul do it.

Actually, you may think about doing the same with Colibrio...

PR #93, opened 2021-11-23
Display must or should

### LINKED ISSUE #74: Additional metadata for MUST and SHOULD?
At the moment, all our tests are related to MUST statements but, eventually, we may also have tests for SHOULD-s. The distinction is significant for reporting: for the CR->PR transition, we will have to report the required number of implementations, whereas no such requirement exists for SHOULD-s. This difference must be clearly noted on the test report and this leads to the necessity to provide this information in the package metadata.

I did not find any obvious way to represent this information in the `dcterms` vocabulary. With a bit of an abuse of the existing vocabularies, I have come up with two different approaches.

One is to look at the tests as a collection in terms of the EPUB metadata vocabulary and add the following:

```xml
    <meta property="belongs-to-collection">must</meta>
```

The value for the `belongs-to-collection` term is `must` or `should`.

If this is considered to be an abuse, we could define our own "vocabulary":

```xml
<package … prefix="testType: https://www.w3.org/ns/test_types/">
...
    <meta property="testType:statement">must</meta>
...
```

again with the value of the terms defined as `must` or `should`.

In both cases the missing metadata value should be taken as a `must` by default (this means we do not require reworking all the current tests...

Any better approach?

Cc: @dlazin @mattgarrish @wareid @dauwhe 

---

As for the UI, my idea is to group the SHOULD tests at the end of the various tables, label all the rows with "M" or "S" (also, possibly display them with a slightly different background color). The point is that it should be clear to, mentally, separate the two category of tests when reviewing the results.

iherman:
This PR implements https://github.com/w3c/epub-tests/issues/74: the tables have a new column with 'required' and 'optional', based on the metadata item in #74:

```
<meta property="belongs-to-collection">should</meta>
```

The tables rows are also reordered by putting the non-required (ie, SHOULD) tests at the bottom of the table.

Here is a visual version of what is generated:

- https://raw.githack.com/w3c/epub-tests/display_must_or_should/docs/drafts/results.html
- https://raw.githack.com/w3c/epub-tests/display_must_or_should/docs/drafts/index.html

Note, in both cases, the section on package documents, and the three `pkg-flow-*` tests that are optional.

@dlazin, if this is merged, there might be other tests that would require that extra metadata. By default, tests are considered as MUST, ie, not need to change most of the existing tests.

Fix #74


docs/contributing.md
```diff
@@ -157,12 +157,14 @@ The package document for each test must contain the following metadata, which is
   report. This should be used if the subject of the test is the value of `dc:title` itself (e.g., testing the base direction
   of the `title` element).
 
-* `dcterms:rights` as part of a `link` element: the rights associated with the test. Except for the rare cases the `href` attribute value should be set to `https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document` (i.e., to the W3C Software and Document Notice and Licence).
+* `belongs-to-collection` (optional, as part of a `meta` element): The value is `should` or `must`, and it specifies whether the test corresponds to a _SHOULD_, respectively a _MUST_, statement in the specification. If the metadata is not provided, or any other value is used, the default `must` value is used.
```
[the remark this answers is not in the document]
dlazin: We should probably also note that a "must not" counts as a must, a "should not" counts as a should, "optional" counts as a may, etc. (Don't need to spell these all out, but want people to understand that they shouldn't be literal.)

docs/contributing.md
```diff
@@ -180,6 +182,7 @@ In this example, only the relevant metadata items are shown:
         <link rel="dcterms:rights" href="https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document"/>
         <link rel="dcterms:rightsHolder" href="https://www.w3.org"/>
         <meta property="dcterms:alternative">Title's base direction set to RTL</meta>
+        <meta property="belongs-to-collection">must</meta>
```
dlazin, line 185: Shouldn't this be omitted in the example? I thought we only wanted people to put this in if it's a should or may.

docs/drafts/results.html
```diff
@@ -73,6 +73,10 @@ <h2>Introduction</h2>
 				<a href="#sec-implementer-list"></a> lists all the implementations with the variant in parenthesis, if applicable.
 			</p>
 
+			<p>
+				The report tables also include a separate column (with the heading "Req?") with flag whether the test is required (i.e., testing a MUST statement) or optional (i.e., testing a SHOULD or MAY statement) for a full implementation conformance.
```
dlazin, line 77: As below, I think this shouldn't be a boolean — we should call it something like Priority or Importance or just MUST (as a header). Or even "Req" with no question mark.

package.json
```diff
@@ -7,6 +7,7 @@
     "test": "tests"
   },
   "scripts": {
+    "debug": "./node_modules/ts-node/dist/bin.js src/main.ts -t",
```
dlazin, line 10: Is this intentional, or leftover from your debugging?

src/lib/data.ts
```diff
@@ -205,7 +214,27 @@ async function get_test_metadata(dir_name: string): Promise<TestData[]> {
...
+
+        const get_array_of_meta_values = (property: string, metadata: any): string[] => {
+            return metadata["meta"].filter((entry:any): boolean => entry["$"].property === property)
+        }
+
+        const get_single_meta_value = (property: string, metadata: any): any => {
+            return metadata.meta.find((entry: any): boolean => entry["$"].property === property)
+        }
+
+        const is_required = (metadata: any): boolean => {
```
dlazin, line 226: I think we should pass through the actual value here (maybe all-capsing it). I would like to see MUST or SHOULD or MAY.

If this were a Google review, I would link to the style-guide section that says "Never use booleans for something that has two states now, but might have more later." I have found this to be very good advice :)

src/lib/data.ts
```diff
@@ -218,26 +247,26 @@ async function get_test_metadata(dir_name: string): Promise<TestData[]> {
             normalizeTags : true,
             explicitArray : true,
         });
-        const metadata = package_js.package.metadata[0]
+        const test_metadata = package_js.package.metadata[0]
     
-        const alternate_title = metadata.meta.find((entry: any): boolean => entry["$"].property === "dcterms:alternative");
-        const final_title = alternate_title === undefined ? get_string_value("dc:title", "(No title)") : alternate_title._;
+        // const alternate_title = get_single_meta_value("dcterms:alternative");
+        // const final_title = alternate_title === undefined ? get_string_value("dc:title", "(No title)") : alternate_title._;
```
dlazin, line 253: These are commented out; looks like you moved them to methods. Delete, I guess? Fix in some way.

src/lib/html.ts
```diff
@@ -122,6 +122,14 @@ const create_one_result_table = (data: ImplementationTable, implementers: Implem
                     },
                 },
             },
+            {
+                td : {
+                    $ : {
+                        class : row.required ? Constants.CLASS_MUST : Constants.CLASS_SHOULD,
```
dlazin, line 128: Can you put this class on the tr instead? That way the CSS can color the row differently for non-required stuff, which'll make it easier to skim.

For example:
tr.should td.pass, tr.may td.pass { background-color: lightgreen; }
tr.should td.fail, tr.may td.fail { background-color: pink; }

dlazin (approved):
Thanks for doing this!

docs/contributing.md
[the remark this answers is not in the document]
iherman: First of all, I have made the system three-way, ie, `must`/`should`/`may`. I have also changed to 

> _MUST_ (or _MUST NOT_) 

for each entry in the description.

docs/contributing.md
iherman: I wanted to keep it, because it may not be clear for everyone how exactly to use this property (it is not frequently used). I have added a remark under the example making it clear that that specific entry could be omitted because it is the default.)

docs/drafts/results.html
iherman: I kept "Req" (without the question mark). The horizontal real estate on the table is limited and a longer text would have required too much (horizontal) space.

package.json
iherman: Intentional. I have a `local_test` directory on my local repo (hidden to git) and this makes it easy to make tests without touching the "real" tests. I would prefer to keep it (for my own convenience).

iherman: I have renamed it to `test` though, to make it clearer.

src/lib/data.ts
iherman: I changed it to use "must", "should", or "may". More exactly, I have defined a literal type that can take one of these three values.

src/lib/data.ts
iherman: Oops. I meant to remove them... thanks.

src/lib/html.ts
iherman: Yeah... I would love if I could. The xmljs package is fairly inflexible, alas!, and I did not find a good way to do it yet. At the moment, I set the "Req" column entries to the `must`/`should`/`may` classes; maybe it will be enough.

iherman: What I did, after all, is to

- set the class attribute on the first column entry instead of the Req column
- use the `td.should ~ td, td.should` CSS selector, which refers to the full line (as proposed on [Slack](https://w3ccommunity.slack.com/archives/G01P2RBAD7T/p1637788081050700)

At the moment, the effect is the "gray" out the corresponding rows, but we can see this later.

iherman:
Merging (see also [Dan's approval on Slack](https://w3ccommunity.slack.com/archives/G01P2RBAD7T/p1637784285050000))

PR #95, opened 2021-11-25
Changing the way OPF files are found

iherman:
Instead of using a fixed name in the file, the script follows the "official" way by reading the `container.xml` file first and extract the file name from there.


iherman:
See indirect approval on  [slack](https://w3ccommunity.slack.com/archives/G01P2RBAD7T/p1637784250049200)

PR #96, opened 2021-12-03
Changes on spec references

iherman:
This is the counterpart of https://github.com/w3c/epub-specs/pull/1951

For two tests:

- the references to the RS spec are removed
- the reference to the Core spec is added

For a third tests (also on manifest) the test is put into the "should" category.

This should only be merged if https://github.com/w3c/epub-specs/pull/1951 is merged.


tests/pkg-manifest-unlisted-resource/OPS/package.opf
```diff
@@ -12,8 +12,9 @@
     <dc:title>Resource not listed in manifest</dc:title>
     <link rel="dcterms:rights" href="https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document"/>
     <link rel="dcterms:rightsHolder" href="https://www.w3.org"/>
+    <meta property="belongs-to-collection">should</meta>
     <meta property="dcterms:isReferencedBy">https://www.w3.org/TR/epub-rs-33/#confreq-rendition-rs-manifest</meta>
-    <meta property="dcterms:modified">2021-01-08T00:00:00Z</meta>
+    <meta property="dcterms:modified">2021-12-03T00:00:00Z</meta>
```
dlazin, line 17: We haven't been updating the dates here; they always match the date above. We could change our strategy, but it seems unnecessarily fussy. I prefer simple and consistent.

tests/pkg-manifest-url/OPS/package.opf
```diff
@@ -12,8 +12,8 @@
     <dc:title>Content in arbitrary directory</dc:title>
     <link rel="dcterms:rights" href="https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document"/>
     <link rel="dcterms:rightsHolder" href="https://www.w3.org"/>
-    <meta property="dcterms:isReferencedBy">https://www.w3.org/TR/epub-rs-33/#confreq-rs-pkg-manifest-url</meta>
-    <meta property="dcterms:modified">2021-01-18T00:00:00Z</meta>
+    <meta property="dcterms:isReferencedBy">https://www.w3.org/TR/epub-33/#confreq-root-url</meta>
+    <meta property="dcterms:modified">2021-12-03T00:00:00Z</meta>
```
dlazin, line 16: Same.

dlazin (approved):
Approved if w3c/epub-specs#1951 is committed.

tests/pkg-manifest-url/OPS/package.opf
```diff
@@ -12,8 +12,8 @@
     <dc:title>Content in arbitrary directory</dc:title>
```
dlazin, line 12: Since this is no longer an RS test, the description on line 8, and on page 1 of the publication, should be updated to describe content behavior (I guess referencing EPUBCheck?).

tests/pkg-relative-url/foo/BAR/baz.opf
```diff
@@ -14,8 +14,7 @@
     <link rel="dcterms:rights" href="https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document"/>
```
dlazin, line 14: Same for the description on line 8 here, and page 1.

tests/pkg-manifest-unlisted-resource/OPS/package.opf
iherman: O.k. I will change it back.

tests/pkg-manifest-url/OPS/package.opf
iherman: Yep.

tests/pkg-manifest-url/OPS/package.opf
iherman: (Copying @mattgarrish here, because that is a sligthly more general editorial issue and also @rdeltour because all this originates in https://github.com/w3c/epub-specs/issues/1918. I believe that came up in another thread.) 

The problem I see is that there is no tight separation between the RS and the Core spec in this case. Issue https://github.com/w3c/epub-specs/issues/1918 came up through the wish to avoid unnecessary redundancy in terms of normative statements between the two specs. But that also means that, for an RS, there are some general requirements about properly interpreting the core content where the details might be normatively stated in the content/core spec rather than specific to the RS. I believe this is unavoidable. This is one of those: the issue of spreading the XHTML file in an arbitrary folder is of course a content format issue (hence epubcheck's) _but_ also an RS issue insofar as the RS must process those references and not rely on some fixed pattern.

From a standard point of view, in my view this is correctly stated in the introductory statement of [§3](https://www.w3.org/TR/epub-rs-33/#sec-pkg-doc-manifest):

> Reading Systems MUST process the [Package Document](https://www.w3.org/TR/epub-33/#sec-package-doc)

And we should try to keep at that. The statements in the RS spell out those parts/details that are relevant for the RS only, but the assumption is that the normative statements in the Core spec are also "honored" by the RS whenever appropriate.

For our testing environment: I would propose to refer to this statement for all the tests in this category. I am sure there will be more cases, including for other similar statements (look at all the yellow lines...)

tests/pkg-relative-url/foo/BAR/baz.opf
iherman: See https://github.com/w3c/epub-tests/pull/96#discussion_r762531400

Issue #53, opened 2021-10-20
Test RS processing of data URLs


m22chan:
This test the normative statement found at:

https://www.w3.org/TR/epub-rs-33/#confreq-rs-data-urls
Reading Systems MUST prevent data URLs [RFC2397] from opening in top-level browsing contexts.

dlazin:
There are now two normative statements here, and I think they both need tests:

> Reading Systems MUST prevent data URLs [RFC2397] from opening in top-level browsing contexts [HTML], except when initiated through a Reading System affordance such as a context menu. If a Reading System does not use a top-level browsing context for Top-level Content Documents, for example if the Top-level Content Document is an SVG, it MUST also prevent data URLs from opening as though they are Top-level Content Documents.

https://w3c.github.io/epub-specs/epub33/rs/#sec-epub-rs-conf-data-urls

m22chan:
Ah, I see. So we need something specifically for:

> If a Reading System does not use a top-level browsing context for Top-level Content Documents, for example if the Top-level Content Document is an SVG, it MUST also prevent data URLs from opening as though they are Top-level Content Documents.

I can make a test book with 2 content documents in the spine.
First content document explains test and contains data: URL version of an SVG.
Second content document contains just text.
Test passes if the RS displays the test book as _two_ top-level content documents - i.e. page 1 and page 2, without inserting a page in between for the SVG.

iherman:
Should have been closed a long time ago by virtue of having merged #54...

PR #98, opened 2021-12-04
Add additional data URL related test

m22chan:
Closes issue #53 by adding an addition test for this assertion:

> If a Reading System does not use a top-level browsing context for Top-level Content Documents, for example if the Top-level Content Document is an SVG, it MUST also prevent data URLs from opening as though they are Top-level Content Documents.

Since there will be 2 tests referencing the same normative statement I also gave the original "top-level browsing context" data URL test a more distinctive ID using the underscore-style naming convention.

(I also realized that I should have been making my commits into a branch inside my fork, rather than on 'main' inside my fork. Oops.)


tests/pub-data-urls_top-level-content/OPS/nav.xhtml
```diff
@@ -0,0 +1,13 @@
+<html xmlns="http://www.w3.org/1999/xhtml" xmlns:epub="http://www.idpf.org/2007/ops" xml:lang="en">
+  <head>
+    <title>Data URL not opening as a Top-level Content Document</title>
```
dlazin, line 3: Nit: sentence case for titles. Applies throughout.

tests/pub-data-urls_top-level-content/OPS/package.opf
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<package xmlns="http://www.idpf.org/2007/opf" xmlns:epub="http://www.idpf.org/2007/ops" version="3.0" xml:lang="en" unique-identifier="pub-id">
+  <metadata xmlns:dc="http://purl.org/dc/elements/1.1/">
+    <dc:coverage>Publication Resources</dc:coverage>
+    <dc:creator>Matthew Chan</dc:creator>
+    <dc:date>2021-12-03</dc:date>
+    <dc:description>
+      The reading system must display an img (where src is a data URL) as part of the Top-level Content
```
dlazin, line 8:
```suggestion
      The reading system must display an img (where src is a data URL) as part of the top-level content
```

tests/pub-data-urls_top-level-content/OPS/package.opf
```diff
@@ -0,0 +1,33 @@
...
+      Document in which it is embedded, rather than opening it as thought it were its own Top-level Content Document.
```
dlazin, line 9:
```suggestion
      document in which it is embedded, rather than opening it as thought it were its own top-level content document.
```

tests/pub-data-urls_top-level-content/OPS/page_001.xhtml
```diff
@@ -0,0 +1,13 @@
+<html xmlns="http://www.w3.org/1999/xhtml" xmlns:epub="http://www.idpf.org/2007/ops" xml:lang="en">
+  <head>
+    <title>Data URL not opening as a Top-level Content Document</title>
+    <meta name="viewport" content="width=900, height=600" />
+    <link rel="stylesheet" href="fixed.css" />
+  </head>
+  <body id="p1">
+    <h1>Data URL not opening as a Top-level Content Document</h1>
+    <p id="text1">The following image is an SVG contained within a <code>data:</code> URL.</p>
+  	<p id="text2">Test passes if you can see the SVG below (instead of on its own page as though it were a Top-level Content Document).</p>
```
dlazin, line 10:
```suggestion
    <p id="text2">Test passes if you can see the SVG below (instead of on its own page as though it were a top-level content document).</p>
```

tests/pub-data-urls_top-level-content/OPS/package.opf
```diff
@@ -0,0 +1,33 @@
...
+    </dc:description>
```
dlazin, line 10: Optional: is there a more explicit way in which you can describe what the expected/aberrant behavior here would be? It feels to me like it would be hard to *see* whether the image is top-level rather than merely happening to reflow to the next page. If you can find a way to clarify this, please put it in both the description and the paragraph on page 1.

dlazin (approved):
Super-nit: you do NOT need to fix this, but if it were me I would just call these _browsing and _top-level; IMO "context/content" is unnecessarily long for the filenames/IDs. But it doesn't really matter.

tests/pub-data-urls_top-level-content/OPS/package.opf
m22chan: Happy holidays, dlazin! (I've been rushing for non-epub related deadlines, so super apologies for delay.)

New commit below cleans up sentence case cleans up some nits.
It also clarifies that the svg is absolutely positioned at the bottom of page 1 in a fixed layout epub. If the svg _does_ appear as top-level content, then the RS has definitely done something unexpected.

m22chan:
Hi @dlazin! Finally getting back into the swing of things this year.
Generally speaking, after your review and my additional commit to address your feedback, should I leave a comment specifically saying that I'm done making changes, and that this PR can be merged?
Because yes, this test is finished!

Thank you!

dlazin:
> Hi @dlazin! Finally getting back into the swing of things this year. Generally speaking, after your review and my additional commit to address your feedback, should I leave a comment specifically saying that I'm done making changes, and that this PR can be merged? Because yes, this test is finished!
> 
> Thank you!

Yes, that'd be great! To be clear, is this ready to merge right now, or not? I forget.

m22chan:
Yes, ready to merge right now. Thank you!

I'll also go ahead and make a PR to add this test to the corresponding data-tests attribute in the spec repo.

PR #100, opened 2021-12-08
`epubReadingSystem` object tests

### LINKED ISSUE #99: `epubReadingSystem` object test
I will try to come up with those.

@dlazin, this is a new category of tests, I believe; we may want to agree on the naming upfront (and not at PR time). My proposal:

- Coverage Label: `Scripting`, to be added between the current labels of `Content Documents` and `Navigation Documents`
- Test names: `scr-epubReadingSystem-*` where '*' will probably say "properties" and "features" (I believe I can get away with two tests, but that is still open)

    The `epubReadingSystem` is a mouthful, but that is the nature of the beast... I was wondering to use something like `ers`, but that would mean trading mouthful to cryptic. I would choose mouthful... WDYT?

iherman:
It is a mega-PR again, apologies to the reviewers. There are several categories of tests:

1. testing whether scripting is available in the first place (those are 'should' tests)
2. testing whether `epubReadingSystem` is implemented with the basic properties
3. testing whether the feature queries are all implemented (there is only one test for all queries, though)

However, some tests have several variations:

- test whether the test is also available when it is part of a content embedded via an iframe
- test whether the spine level content, or the one embedded via an iframe, is in SVG rather than XHTML

Fixes #99


dlazin (commented):
As I mentioned on https://github.com/w3c/epub-specs/pull/1956, I would — surprise, sorry — like to rename these.

a) three-letter abbreviation for consistency with other tests (scr-?)
b) all lowercase

So, for example, #scr-readingsystem-features, etc?

Please rename and then I will review. Thanks!

iherman:
@dlazin I have made the name changes

tests/scr-readingsystem-features/OPS/content_001.xhtml
```diff
@@ -0,0 +1,88 @@
...
+      <ul>
+        <li><code>dom-manipulation</code>: <code id="dom-manipulation"></code></li>
+        <li><code>layout-changes</code>: <code id="layout-changes"></code></li>
+        <li><code>touch-events</code>: <code id="touch-events"></code></li>
+        <li><code>mouse-events</code>: <code id="mouse-events"></code></li>
+        <li><code>keyboard-events</code>: <code id="keyboard-events"></code></li>
+        <li><code>spine-scripting</code>: <code id="spine-scripting"></code></li>
+      </ul>
+      <p>
+        If all entries are <code>true</code> or <code>false</code>, the test passes.
```
dlazin, line 82: Nit: "false" is slightly unclear as a pass condition. I know this means "epubReadingSystem exists but the environment does not support, say, touch-events," but I think this would be clearer as something like enabled/disabled/unimplemented, with "unimplemented" as the test failure condition.

(Enabled/disabled doesn't seem like quite the right terminology either, but I'm blanking on the best word.)

Optional, though. As a minimal change, I would suggest just changing n/a to "unimplemented," because that's more explicit than n/a.

tests/scr-readingsystem-features/OPS/package.opf
```diff
@@ -0,0 +1,25 @@
...
+    <dc:creator>Ivan Herman</dc:creator>
+    <dc:date>2021-12-10</dc:date>
+    <dc:description>Checks whether the reading system implements the required epubReadingSystem feature string queries.</dc:description>
+    <dc:identifier id="pub-id">scr-readingsystem-features</dc:identifier>
+    <dc:language>en</dc:language>
+    <dc:title>epubReadingSystem object features</dc:title>
+    <link rel="dcterms:rights" href="https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document"/>
+    <link rel="dcterms:rightsHolder" href="https://www.w3.org"/>
+    <meta property="dcterms:isReferencedBy">https://www.w3.org/TR/epub-rs-33/#scripting-req-readingsystem</meta>
+    <meta property="dcterms:isReferencedBy">https://www.w3.org/TR/epub-rs-33/#scripting-req-readingsystem-features</meta>
```
dlazin, line 14: These two IDs don't exist. Is that because you're adding them in a follow PR? Fine if so.

tests/scr-readingsystem-support/OPS/content_001.xhtml
```diff
@@ -0,0 +1,58 @@
...
+
+    <div id="fail" style="display: block">
+      <p>The Reading System does not implement the <code>epubReadingSystem</code> object. Test fails.</p>
+    </div>
+
+    <div id="pass" style="display: none">
+      <p>The Reading System implements the <code>epubReadingSystem</code> object, with the following properties:</p>
+      <ul>
+        <li>Reading System Name: <span id="ers_name"></span></li>
+        <li>Reading System Version: <span id="ers_version"></span></li>
```
dlazin, line 52: If ERS exists but either or neither of these is populated, does the test still pass? Please clarify.

tests/scr-readingsystem-support/OPS/package.opf
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<package xmlns="http://www.idpf.org/2007/opf" xmlns:epub="http://www.idpf.org/2007/ops" version="3.0" xml:lang="en" unique-identifier="pub-id">
+  <metadata xmlns:dc="http://purl.org/dc/elements/1.1/">
+    <dc:coverage>Scripting</dc:coverage>
+    <dc:creator>Ivan Herman</dc:creator>
+    <dc:date>2021-12-10</dc:date>
+    <dc:description>Checks whether the reading system implements the epubReadingSystem object in a spine-level document.</dc:description>
```
dlazin, line 7: What does "in a spine-level document" imply here? As opposed to what? I recommend either removing those words or clarifying how this differs from what it *would* mean if those words didn't exist.

I think you mean "as opposed to the iframe test" but I can't imagine an RS implementing ERS in an iframe but *not* in the main document, so I would tend to omit the extra language here, implying that this is the base case, and then have the iframe test as is (essentially, a more specific test).

tests/scr-readingsystem-support_iframe/OPS/content_001.xhtml
```diff
@@ -0,0 +1,26 @@
...
+
+      }
+
+      window.addEventListener('load', test_ers);
+    </script>
+  </head>
+  <body>
+    <p id="scripting_support">This Reading System does not support scripting.</p>
+
+    <iframe src="iframe_content.xhtml" width="600" height="600" style="border: solid;" title="Embedded iframe to check the availability of the epubReadingSystem object support"></iframe>
```
dlazin, line 23: Is it possible that some reading systems do not support iframes at all, and thus this test would end up in an ambiguous state with no clear pass/fail message? If so I think the previous paragraph should note that there's an iframe below, and if it doesn't say it passes, the test fails.

tests/scr-readingsystem-support_iframe/OPS/iframe_content.xhtml
```diff
@@ -0,0 +1,53 @@
...
+  <body>
+    <div id="fail" style="display: block">
+      <p>The Reading System does not implement the <code>epubReadingSystem</code> object within an <code>iframe</code>. Test fails.</p>
+    </div>
+
+    <div id="pass" style="display: none">
+      <p>The Reading System implements the <code>epubReadingSystem</code> object within an <code>iframe</code>, with the following properties:</p>
+      <ul>
+        <li>Reading System Name: <span id="ers_name"></span></li>
+        <li>Reading System Version: <span id="ers_version"></span></li>
```
dlazin, line 47: Same question as previous test: what if ERS is defined but one or both of these properties is not?

tests/scr-readingsystem-features/OPS/content_001.xhtml
```diff
@@ -0,0 +1,88 @@
...
+
+      /**
+       * A simple change that proves that scripting is enabled.
+       * 
+       * Note that the script is more complicated than would be necessary in HTML, because the innerHTML method does not
+       * seem to work in xhtml (only in html) :-(
+       */ 
+      function test_ers_features() {
+        const support = document.getElementById('scripting_support');
+        support.textContent = 'This Reading System supports scripting.';
```
dlazin, line 22: Nit: we usually don't capitalize Reading System. (For example, you did not in your description tags.) Applies throughout PR, but optional.

tests/scr-readingsystem-support_iframe/OPS/package.opf
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<package xmlns="http://www.idpf.org/2007/opf" xmlns:epub="http://www.idpf.org/2007/ops" version="3.0" xml:lang="en" unique-identifier="pub-id">
+  <metadata xmlns:dc="http://purl.org/dc/elements/1.1/">
+    <dc:coverage>Scripting</dc:coverage>
+    <dc:creator>Ivan Herman</dc:creator>
+    <dc:date>2021-12-10</dc:date>
+    <dc:description>Checks whether the reading system implements the epubReadingSystem object in container-constraint document.</dc:description>
```
dlazin, line 7: Should this be container-constrained? Maybe this is a term of art, but in normal English constraint isn't an adjective or adverb. Also needs either "a" or "documents" plural.

Also, I think this should say something specifically about an iframe, like "in a container-constrained document (e.g. an iframe)."

tests/scr-readingsystem-support_iframe_svg/OPS/content_001.xhtml
```diff
@@ -0,0 +1,26 @@
+<html xmlns="http://www.w3.org/1999/xhtml" xmlns:epub="http://www.idpf.org/2007/ops" xml:lang="en">
+  <head>
+    <title>epubReadingSystem object support in SVG in an iframe</title>
```
dlazin, line 3: Nit: "in an SVG". Applies throughout.

tests/scr-readingsystem-support_iframe_svg/OPS/iframe_content.svg
```diff
@@ -0,0 +1,80 @@
+<svg version="1.1" viewBox="0 0 300 200" xmlns="http://www.w3.org/2000/svg">
+  <script>//<![CDATA[
```
dlazin, line 2: Why do only 3 of these files have CDATA? Fine if appropriate, but I'm unfamiliar with the need/distinction.

dlazin (commented):
Great tests! I don't have enough time right now, so this is just a partial review — more later.

tests/scr-readingsystem-features/OPS/content_001.xhtml
```diff
@@ -0,0 +1,88 @@
...
+  <head>
+    <title>epubReadingSystem object features</title>
+    <script>
+      const required_features = [
+        'dom-manipulation', 
+        'layout-changes', 
+        'touch-events', 
+        'mouse-events',
+        'keyboard-events',
+        'spine-scripting'
```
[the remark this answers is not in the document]
iherman: It follows the same order as in the [spec](https://www.w3.org/TR/epub-rs-33/#app-ers-hasFeature). I would prefer to keep it that way.

tests/scr-readingsystem-features/OPS/content_001.xhtml
iherman: +1

tests/scr-readingsystem-features/OPS/content_001.xhtml
iherman: Agree with 'unimplemented'. As for the other values: the [spec](https://www.w3.org/TR/epub-rs-33/#app-ers-hasFeature) defines these methods as returning a boolean value, ie, I think the best is to keep it as is.

tests/scr-readingsystem-features/OPS/content_001.xhtml
iherman: I also made the changes in the other tests

tests/scr-readingsystem-features/OPS/package.opf
iherman: Yep, they are in https://github.com/w3c/epub-specs/pull/1956

tests/scr-readingsystem-support/OPS/content_001.xhtml
iherman: Yep. It is unlikely that an implementation implements the object without the only two attributes that are defined, but let us be complete. I have modified the description by:

> `<p>Test passes, unless one of the two values are set to <code>undefined</code>.</p>`

tests/scr-readingsystem-support/OPS/package.opf
iherman: Agreed.

tests/scr-readingsystem-support/OPS/content_001.xhtml
```diff
@@ -0,0 +1,58 @@
...
+      <p>The reading system does not implement the <code>epubReadingSystem</code> object. Test fails.</p>
+    </div>
+
+    <div id="pass" style="display: none">
+      <p>The reading system implements the <code>epubReadingSystem</code> object, with the following properties:</p>
+      <ul>
+        <li>Reading System Name: <span id="ers_name"></span></li>
+        <li>Reading System Version: <span id="ers_version"></span></li>
+      </ul>
+      <p>Test passes, unless one of the two values are set to <code>undefined</code>.</p>
```
dlazin, line 54: Nit: I think this should be "either" rather than "one of". Both of them being undefined would also be a failure, and "either" implies that.

tests/scr-readingsystem-support_iframe/OPS/content_001.xhtml
iherman: Well... Our basic assumption in our testing environments is that a bona fide reading system implements HTML5, and iframe is, afaik, a required part of it. But I have added an additional sentence into, please look at this (and we may want to replicate this to all tests relying on iframes):

> `<p>Note that the test relies on the availability of the <code>iframe</code> element in HTML. If no content appears above this line, this means the <code>iframe</code> is not available, and the test fails.</p>`

tests/scr-readingsystem-support_iframe_svg/OPS/package.opf
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<package xmlns="http://www.idpf.org/2007/opf" xmlns:epub="http://www.idpf.org/2007/ops" version="3.0" xml:lang="en" unique-identifier="pub-id">
+  <metadata xmlns:dc="http://purl.org/dc/elements/1.1/">
+    <dc:coverage>Scripting</dc:coverage>
+    <dc:creator>Ivan Herman</dc:creator>
+    <dc:date>2021-12-10</dc:date>
+    <dc:description>Checks whether the reading system implements the epubReadingSystem object in container-constraint document.</dc:description>
```
dlazin, line 7: This seems to be the same description as scr-readingsystem-support_iframe/OPS/package.opf.

tests/scr-readingsystem-support_iframe/OPS/iframe_content.xhtml
iherman: Right, and I have copied the same text as in https://github.com/w3c/epub-tests/pull/100#discussion_r783149590

tests/scr-readingsystem-support_iframe/OPS/package.opf
iherman: Yep, my mistake. Done.

tests/scr-readingsystem-support_iframe_svg/OPS/content_001.xhtml
iherman: +1

dlazin (commented):
Slightly more!

tests/scr-readingsystem-support/OPS/content_001.xhtml
iherman: +1

tests/scr-readingsystem-support_iframe_svg/OPS/package.opf
iherman: This has changed under your feet:-) I think it is better if I stop now (it is late afternoon anyway) and pick this up tomorrow or wherever you did your next batch :-)

tests/scr-readingsystem-support_iframe_svg/OPS/iframe_content.svg
dlazin: (Oh, because they're SVGs!)

tests/scr-readingsystem-support_iframe_svg/OPS/iframe_content.svg
```diff
@@ -0,0 +1,80 @@
...
+    </g>
+  </g>
+
+  <g transform="translate(0, 50)" style="display:none" id="ers_passes">
+    <ellipse cx="150" cy="100" rx="60" ry="30" fill="green" />
+    <g transform="translate(0,-38)">
+      <text x="150" y="125">The reading system</text>
+      <text x="150" y="132">implements the <tspan class="ers">epubReadingSystem</tspan> object.</text>
+      <text x="150" y="139">Reading System Name: <tspan id="ers_name">sg</tspan></text>
+      <text x="150" y="146">Reading System Version: <tspan id="ers_version">1.0.0</tspan></text>
```
dlazin, line 75: Why do these two lines have hardcoded values?

tests/scr-readingsystem-support_svg/OPS/content_001.svg
```diff
@@ -0,0 +1,80 @@
...
+    </g>
+  </g>
+
+  <g transform="translate(0, 50)" style="display:none" id="ers_passes">
+    <ellipse cx="150" cy="100" rx="60" ry="30" fill="green" />
+    <g transform="translate(0,-38)">
+      <text x="150" y="125">The reading system</text>
+      <text x="150" y="132">implements the <tspan class="ers">epubReadingSystem</tspan> object.</text>
+      <text x="150" y="139">Reading System Name: <tspan id="ers_name">sg</tspan></text>
+      <text x="150" y="146">Reading System Version: <tspan id="ers_version">1.0.0</tspan></text>
```
dlazin, line 75: Same question about hardcoded values.

tests/scr-support_iframe/OPS/package.opf
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<package xmlns="http://www.idpf.org/2007/opf" xmlns:epub="http://www.idpf.org/2007/ops" version="3.0" xml:lang="en" unique-identifier="pub-id">
+  <metadata xmlns:dc="http://purl.org/dc/elements/1.1/">
+    <dc:coverage>Scripting</dc:coverage>
+    <dc:creator>Ivan Herman</dc:creator>
+    <dc:date>2021-12-10</dc:date>
+    <dc:description>Checks whether the reading system implements scripting in a container-constrained document (i.e., whether a simple inline script is executed at all).</dc:description>
+    <dc:identifier id="pub-id">scr-support_iframe</dc:identifier>
+    <dc:language>en</dc:language>
+    <dc:title>Scripting support</dc:title>
```
dlazin, line 10: ...in an iframe? (Applies to other <titles> in the same test too.)

dlazin (approved):
All done! Thanks for these.

tests/scr-readingsystem-support_iframe_svg/OPS/iframe_content.svg
iherman: These were just phony values that were used _testing_ the SVG. The enclosed `<g>` is, by default, not displayed; it is only displayed if scripting is enabled in the first place and the values are changed by the real ones (see lines 26 and 27 in the script). But, just to avoid any problems, I have removed these.

tests/scr-readingsystem-support_svg/OPS/content_001.svg
iherman: With the same answer :-)

tests/scr-support_iframe/OPS/package.opf
iherman: I have actually changed the text to 

> Checks whether the reading system implements container-constrained scripts (i.e., whether a simple inline script is executed at all in an iframe).

Per [content spec](https://www.w3.org/TR/epub-33/#sec-scripted-container-constrained) "container-constrained script" is the official statement. (Will do in all other tests.)

PR #112, opened 2022-01-25
Adding a navigation in spine test: the list style must not affect the way TOC is displayed

### LINKED ISSUE #111: List item numbering style not affecting TOC rendering
Testing the statement:

> MUST NOT show list item numbering on nav elements when presenting them outside of the spine, regardless of support for CSS

See https://w3c.github.io/epub-specs/epub33/rs/#sec-nav

iherman:
The new test is `nav-spine_in-spine-no-list-style`. On the side, the package documents for all other nav related tests have been updated; the references to the RS documents were wrong.

Fix #111


tests/nav-spine_in-spine-no-list-style/OPS/content_001.xhtml
```diff
@@ -0,0 +1,8 @@
+<html xmlns="http://www.w3.org/1999/xhtml" xmlns:epub="http://www.idpf.org/2007/ops" xml:lang="en">
+  <head>
+    <title>TOC items ignore navigation list styling</title>
+  </head>
+  <body>
+    <p>Test passes if you can access the two links in the navigation document, which is in the spine.</p>
```
dlazin, line 6: I don't think this is the correct description.

tests/nav-spine_in-spine-no-list-style/OPS/package.opf
```diff
@@ -0,0 +1,29 @@
...
+    <meta property="dcterms:isReferencedBy">https://www.w3.org/TR/epub-rs-33/#confreq-nav-ol-style</meta>
+    <meta property="dcterms:modified">2022-01-25T00:00:00Z</meta>
+  </metadata>
+  <manifest>
+    <item id="content_001" href="content_001.xhtml" media-type="application/xhtml+xml" />
+    <item id="content_002" href="content_002.xhtml" media-type="application/xhtml+xml" />
+    <item id="nav" properties="nav" href="nav.xhtml" media-type="application/xhtml+xml" />
+  </manifest>
+  <spine>
+    <itemref idref="nav" />
```
dlazin, line 25: Is there a need for the nav to be in the spine? I think it confuses the test, because you'll see it once as content and once as UI, and the test only applies to the UI part (and the content part is likely to "fail" the test). Remove?

tests/nav-spine_in-spine-no-list-style/OPS/package.opf
```diff
@@ -0,0 +1,29 @@
...
+<package xmlns="http://www.idpf.org/2007/opf" xmlns:epub="http://www.idpf.org/2007/ops" version="3.0" xml:lang="en" unique-identifier="pub-id">
+  <metadata xmlns:dc="http://purl.org/dc/elements/1.1/">
+    <dc:coverage>Navigation Documents</dc:coverage>
+    <dc:creator>Ivan Herman</dc:creator>
+    <dc:creator>Dave Cramer</dc:creator>
+    <dc:date>2022-01-25</dc:date>
+    <dc:description>
+      The navigation document is also referenced in the spine; however, the TOC items in the Reading System's own control must not use the list style type of the navigation document.
+    </dc:description>
+    <dc:identifier id="pub-id">nav-spine_in-spine-no-list-style</dc:identifier>
```
dlazin, line 11: As below, I don't see a need for the nav to be in the spine in this test. I would rename this test to nav-list-style.

tests/nav-spine_in-spine-no-list-style/OPS/package.opf
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<package xmlns="http://www.idpf.org/2007/opf" xmlns:epub="http://www.idpf.org/2007/ops" version="3.0" xml:lang="en" unique-identifier="pub-id">
+  <metadata xmlns:dc="http://purl.org/dc/elements/1.1/">
+    <dc:coverage>Navigation Documents</dc:coverage>
+    <dc:creator>Ivan Herman</dc:creator>
+    <dc:creator>Dave Cramer</dc:creator>
+    <dc:date>2022-01-25</dc:date>
+    <dc:description>
+      The navigation document is also referenced in the spine; however, the TOC items in the Reading System's own control must not use the list style type of the navigation document.
```
dlazin, line 9:
```suggestion
      The reading system's rendering of the navigation must not show numbering, disregarding the list-style-type declared in the navigation document.
```

tests/nav-spine_in-spine-no-list-style/OPS/package.opf
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<package xmlns="http://www.idpf.org/2007/opf" xmlns:epub="http://www.idpf.org/2007/ops" version="3.0" xml:lang="en" unique-identifier="pub-id">
+  <metadata xmlns:dc="http://purl.org/dc/elements/1.1/">
+    <dc:coverage>Navigation Documents</dc:coverage>
+    <dc:creator>Ivan Herman</dc:creator>
+    <dc:creator>Dave Cramer</dc:creator>
```
dlazin, line 6: Why is Dave in this one? I suspect you branched from one of his tests, but in that case he's an author of every test :)

dlazin (approved):
Sorry for slow review!

tests/nav-spine_in-spine-no-list-style/OPS/package.opf
iherman: ok

tests/nav-spine_in-spine-no-list-style/OPS/package.opf
iherman: +1

tests/nav-spine_in-spine-no-list-style/OPS/content_001.xhtml
iherman: I have taken over your text in the package document, with a slight change.

tests/nav-spine_in-spine-no-list-style/OPS/package.opf
iherman: I respectfully disagree on this one. 

In general, of course, there is no reason to put the nav into the spine (unless the nav file contains other things). But the very essence of the test is to show that one can add all kinds of styling to the navigation elements that are perfectly fine CSS statement that the Reading System can use when rendering the nav file as HTML but, on the other hand, the same styling is ignored when the navigation elements are displayed in the Reading System's chrome. To show that this navigation uses a unusual numbering of the (latin) text, ie, lower case greek characters.

I.e., the tests applies on the UI part _as opposed to the HTML styling_. Hence having the nav file in the spine.

tests/nav-spine_in-spine-no-list-style/OPS/package.opf
iherman: See the discussion below.

dlazin: OK, but even so I think this needs a rename. The current style, nav-spine_something, means this test relates to a conformance requirement that *could* be called nav-spine, but we've already used that name for a separate requirement. Therefore this should still be nav-list-style, or nav-list-style_in-spine if you were going to have an in-spine/not-in-spine pair of tests (which seems unnecessary).

tests/nav-spine_in-spine-no-list-style/OPS/package.opf
dlazin: OK, that's fair, but in that case I think the text in nav.xhtml needs to be made clearer. It should basically say "doesn't matter what you see in the 'content area' [whatever we call that], although if the RS supports CSS it will probably have lower-greek numbering; look only at the navigation in the reading system UI."

dlazin (approved):

PR #113, opened 2022-01-28
Getting rid of xml2js in the html.ts file

iherman:
The usage of the `xml2js` to generate the various HTML fragments (in `html.ts`) turned out to be overly complex and very difficult to read (i.e., to maintain). This PR re-writes `html.ts` using the `jsdom` package.

As an aside, the output of the script has been slightly improved, mainly for the display of variants.


src/lib/html.ts
```diff
@@ -49,374 +43,177 @@ const convert_to_id = (header: string): string => {
...
-            return name
-        }
-    }); 
+    const ol = add_child(section, 'ol');
 
-    // This is where the object is turned into an XML serialization...
-    return builder.buildObject(root)
+    for (const implementer of impl) {
+        const li = add_child(ol, 'li');
+        const name = 'ref' in implementer ? `<a href=${implementer.ref}>${implementer.name}</a>` : `${implementer.name}`;
```
dlazin, line 53:
```suggestion
        const name = 'ref' in implementer ? `<a href="${implementer.ref}">${implementer.name}</a>` : `${implementer.name}`;
```

dlazin: Also, note (as I'm sure you know) that constructing HTML like this is not generally considered safe, but given the scope of this script I think it's OK.

dlazin (approved):
Sorry for the slow review!

PR #116, opened 2022-02-09
Additional MO tests

iherman:
Adding four new tests related to [§8.2](https://w3c.github.io/epub-specs/epub33/rs/#sec-behaviors-playback) (plus minor update to the test submitted in the previous PR).

Technically, the new tests do the following (describing it here to make review easier, and also to see if my understanding of the spec is indeed o.k.)

- `mol-css_classes`: this is more-or-less a copy from the spec, using an embedded css for the display of the basic content. Otherwise it is identical to `mo-timing-synchronization`.
- `mol-audio_no_clipbegin`: the `clipBegin` from the very first entry in the MO file is removed. That entry refers, usually, to the place in the mobydick clip where the real narrative starts; in this case the read should start by the 'preamble' (title, etc).
- `mol-audio_no_clipend`: the `clipEnd` from the last entry in the MO file is removed. The MO file only defines a `par` for the first two `<span>` entries in the HTML file; the effect of removing `clipEnd` is that the reading continues until the end of the audio file, ie, includes the segment that is not referred to in the MO file (but the CSS highlight, if rendered, stops at the second fragment).
- `mol-audio_exceeding_clipend`: the mo file refers to _two_ different audio files. The first three `par` entries refer to the first audio file, and the fourth refers to the second audio file. The `clipEnd` value of the third `par` (which refers to the first audio file) exceeds the length of the first audio file. Per spec the net effect is that the reading should continue immediately to the second audio file, essentially ignoring the excess time given in the `clipEnd` of the third `par`.

P.S. Interestingly, `mol-audio-exceeding_clipend` is also a good test for the situation where the mo smil file refers to _several_ audio files. This feature seems to break on the current version of the Thorium reader, but it works on the Web version of Colibrio. I did not find any place in the spec that explicitly says that using several audio files referred from a smil file is o.k., though it seems obvious. Maybe it is worth explicitly mentioning it (and attach `mol-audio-exceeding_clipend` as a test)?


tests/mol-audio_exceeding_clipend/META-INF/container.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
dlazin, line 1: SORRRRRRY but test names should have only one underscore (it's a separator between the hyphenated indicator of what requirement we're testing and a descriptor of the test, also possibly hyphenated but usually just one word).

Based on the (unsubmitted) anchor being https://www.w3.org/TR/epub-rs-33/#mol-audio-exceeding-clipEnd, which seems to correspond to "If clipEnd exceeds the full duration of the physical media, Reading Systems MUST assume its value to be the full duration of the physical media," I think this should not be hyphenated at all. I think it should be mol-audio-exceeding-clipend. OK?

(I guess I really haven't explained this properly, but the format is mol-some-identifier-for-the-normative-statement_modifier-if-necessary. Sort of like BEM in CSS (https://css-tricks.com/bem-101/), if you think of the three-letter abbreviation (mol-) as the Block, the normative statement name as the Element, and the optional modifier as the Modifier.)

tests/mol-audio_exceeding_clipend/OPS/content_001.xhtml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<html xmlns="http://www.w3.org/1999/xhtml">
+	<head>
+        <meta charset="utf-8" />
+		<title>Ignoring clipEnd attribute if it exceeds track length</title>
+	</head>
```
dlazin, line 6: Same problem as #118 with mix of tabs and spaces.

tests/mol-audio_exceeding_clipend/OPS/content_001.xhtml
```diff
@@ -0,0 +1,17 @@
...
+	<body>
+
+        <p>Test passes (i.e., the Reading System correctly implements the <code>clipEnd</code> value exceeding the end of the audio track) if</p>
```
dlazin, line 9:
```suggestion
        <p>Test passes (i.e., the Reading System correctly handles the <code>clipEnd</code> value exceeding the end of the audio track) if</p>
```

tests/mol-audio_exceeding_clipend/OPS/content_001.xhtml
```diff
@@ -0,0 +1,17 @@
...
+
+        <ol>
+            <li>The Reading System provides the user with a way to turn text reading on</li>
+            <li>If turned on, the extract on the next page is read aloud by the Reading System</li>
+            <li>The full text is read without a significant gap between the first and the second paragraph</li>
```
dlazin, line 14: Period at end? Optional but consistent with #118, please.

tests/mol-audio_exceeding_clipend/OPS/mobydick.xhtml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="utf-8"?>
+<html xmlns="http://www.w3.org/1999/xhtml">
+	<head>
+        <meta charset="utf-8" />
+		<title>Ignoring clipEnd attribute if it exceeds track length</title>
```
dlazin, line 5: Inconsistent indentation again.

tests/mol-audio_exceeding_clipend/OPS/package.opf
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<package xmlns="http://www.idpf.org/2007/opf" xmlns:epub="http://www.idpf.org/2007/ops" version="3.0" xml:lang="en" unique-identifier="pub-id">
+  <metadata xmlns:dc="http://purl.org/dc/elements/1.1/">
+    <dc:coverage>Media Overlays</dc:coverage>
+    <dc:creator>Ivan Herman</dc:creator>
+    <dc:creator>Marisa DeMeglio</dc:creator>
+    <dc:date>2022-02-10</dc:date>
+    <dc:description>Tests whether the Reading System ignores the value of the clipEnd attribute in the smil file's 'par' element if it exceeds the length of the audio file, and moves to the next 'par' step immediately at the end that audio content</dc:description>
```
dlazin, line 8:
```suggestion
    <dc:description>Tests whether the Reading System ignores the value of the clipEnd attribute in the smil file's 'par' element if it exceeds the length of the audio file, and moves to the next 'par' step immediately at the end that audio content.</dc:description>
```

dlazin: Optionally cut everything after the comma. Seems like too much detail for the test description, to me.

tests/mol-audio_exceeding_clipend/OPS/package.opf
```diff
@@ -0,0 +1,34 @@
...
+    <dc:identifier id="pub-id">mol-exceeding_clipend</dc:identifier>
+    <dc:language>en</dc:language>
+    <dc:title>Ignoring clipEnd attribute if it exceeds track length</dc:title>
+    <link rel="dcterms:rights" href="https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document"/>
+    <link rel="dcterms:rightsHolder" href="https://www.w3.org"/>
+    <meta property="belongs-to-collection">may</meta>
+    <meta property="dcterms:isReferencedBy">https://www.w3.org/TR/epub-rs-33/#mol-audio-exceeding-clipEnd</meta>
```
dlazin, line 15: I think this is unsubmitted, but for consistency with existing spec anchors (and the naming of this test), I think this anchor should be -clipend, not -clipEnd.

tests/mol-audio_no_clipbegin/OPS/content_001.xhtml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<html xmlns="http://www.w3.org/1999/xhtml">
+	<head>
+        <meta charset="utf-8" />
+		<title>Automatic start at the beginning of an audio track</title>
```
dlazin, line 5: Inconsistent indentation again. (Should there be an xx-mol-template folder, like https://github.com/w3c/epub-tests/tree/main/tests/xx-epub-template, and if so can you make sure the template is tidy?)

tests/mol-audio_no_clipbegin/OPS/content_001.xhtml
```diff
@@ -0,0 +1,18 @@
...
+	</head>
+	<body>
+
+        <p>Test passes (i.e., the Reading System correctly implements the lack of a <code>clipBegin</code> attribute value) if</p>
+
+        <ol>
+            <li>The Reading System provides the user with a way to turn text reading on</li>
+            <li>If turned on, the extract on the next page is read aloud by the Reading System</li>
+            <li>The text does <em>not</em> start at the beginning of the text but at the title of the publication, name of the reader, etc.</li>
```
dlazin, line 14: Should "The text does not start" be "The audio does not start" or similar? Not a Media Overlays expert but "text" seems to me like the wrong word.

tests/mol-audio_no_clipbegin/OPS/package.opf
```diff
@@ -0,0 +1,33 @@
...
+<package xmlns="http://www.idpf.org/2007/opf" xmlns:epub="http://www.idpf.org/2007/ops" version="3.0" xml:lang="en" unique-identifier="pub-id">
+  <metadata xmlns:dc="http://purl.org/dc/elements/1.1/">
+    <dc:coverage>Media Overlays</dc:coverage>
+    <dc:creator>Ivan Herman</dc:creator>
+    <dc:creator>Marisa DeMeglio</dc:creator>
+    <dc:date>2022-02-10</dc:date>
+    <dc:description>Tests whether the Reading System starts at the beginning of the audio track in the absence of a clipBegin attribute in the smil file</dc:description>
+    <dc:identifier id="pub-id">mol-audio_no_clipbegin</dc:identifier>
+    <dc:language>en</dc:language>
+    <dc:title>Automatic start at the beginning of an audio track</dc:title>
```
dlazin, line 11: Optional, but consider making the title reference clipBegin to match the clipEnd test above.

tests/mol-audio_no_clipend/OPS/content_001.xhtml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<html xmlns="http://www.w3.org/1999/xhtml">
+	<head>
+        <meta charset="utf-8" />
+		<title>Automatic rendering of an audio track until the end</title>
+	</head>
+	<body>
+
+        <p>Test passes (i.e., the Reading System correctly implements the lack of a <code>clipEnd</code> attribute value) if</p>
```
dlazin, line 9:
```suggestion
        <p>Test passes (i.e., the Reading System correctly handles the lack of a <code>clipEnd</code> attribute) if</p>
```

tests/mol-audio_no_clipbegin/OPS/content_001.xhtml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<html xmlns="http://www.w3.org/1999/xhtml">
+	<head>
+        <meta charset="utf-8" />
+		<title>Automatic start at the beginning of an audio track</title>
+	</head>
+	<body>
+
+        <p>Test passes (i.e., the Reading System correctly implements the lack of a <code>clipBegin</code> attribute value) if</p>
```
dlazin, line 9:
```suggestion
        <p>Test passes (i.e., the Reading System correctly handles the lack of a <code>clipBegin</code> attribute) if</p>
```

tests/mol-audio_no_clipbegin/META-INF/container.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
dlazin, line 1: mol-audio-no-clipbegin, please :)

tests/mol-css_classes/OPS/mo/mobydick.smil
```diff
@@ -0,0 +1,65 @@
...
+    <body>
+        <seq epub:textref="../mobydick.xhtml#mobyexcerpt">
+            <par id="word1">
+                <text src="../mobydick.xhtml#c01w00001"/>
+                <audio src="../audio/mobydick.mp4" clipBegin="0:00:29.268" clipEnd="0:00:29.441"/>
+            </par>
+
+            <par id="word2">
+                <text src="../mobydick.xhtml#c01w00002"/>
+                <audio src="../audio/mobydick.mp4" clipBegin="0:00:29.441" clipEnd="0:00:29.640"/>
```
dlazin, line 11: Consider making some of these have no clipEnd / no clipBegin / bad clipEnd to see whether that interacts correctly with CSS.

tests/mol-css_classes/OPS/mobydick.xhtml
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="utf-8"?>
+<html xmlns="http://www.w3.org/1999/xhtml">
+	<head>
+        <meta charset="utf-8" />
+		<title>Basic media overlay test with CSS styling</title>
```
dlazin, line 5: Indentation.

tests/mol-css_classes/OPS/package.opf
```diff
@@ -0,0 +1,35 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<package xmlns="http://www.idpf.org/2007/opf" xmlns:epub="http://www.idpf.org/2007/ops" version="3.0" xml:lang="en" unique-identifier="pub-id">
+  <metadata xmlns:dc="http://purl.org/dc/elements/1.1/">
+    <dc:coverage>Media Overlays</dc:coverage>
+    <dc:creator>Ivan Herman</dc:creator>
+    <dc:creator>Marisa DeMeglio</dc:creator>
+    <dc:date>2022-02-10</dc:date>
+    <dc:description>Tests whether the Reading System implements the associated CSS styling for media overlays</dc:description>
```
dlazin, line 8:
```suggestion
    <dc:description>Tests whether the Reading System implements CSS styling for media overlays.</dc:description>
```

dlazin:
> I did not find any place in the spec that explicitly says that using several audio files referred from a smil file is o.k., though it seems obvious. Maybe it is worth explicitly mentioning it (and attach `mol-audio-exceeding_clipend` as a test)?

Yes, I think you should file a spec bug for this, and assuming we agree it should be in the spec, then you can attach that test.

dlazin (approved):
Sorry for so many comments, but most are very easy / repetitive. Thanks again!

tests/mol-audio_exceeding_clipend/OPS/content_001.xhtml
iherman: +1

tests/mol-audio_exceeding_clipend/OPS/content_001.xhtml
iherman: I chose no period in #118...

tests/mol-audio_no_clipbegin/OPS/content_001.xhtml
iherman: The problem with a template is that there are many factors that make the tests fairly different from one another (smil files, audio files, etc). Let us come back to this issue when we are done with the tests.

tests/mol-audio_no_clipbegin/OPS/package.opf
iherman: Chaned

tests/mol-audio_no_clipbegin/OPS/package.opf
```diff
@@ -0,0 +1,33 @@
...
+    <link rel="dcterms:rights" href="https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document"/>
+    <link rel="dcterms:rightsHolder" href="https://www.w3.org"/>
+    <meta property="belongs-to-collection">may</meta>
```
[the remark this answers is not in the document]
iherman: Looking at the issue discussion I think this will become a should overall, and it is an important a11y feature that we should push for...

tests/mol-audio_no_clipend/OPS/content_001.xhtml
iherman: +1

tests/mol-audio_no_clipend/OPS/content_001.xhtml
```diff
@@ -0,0 +1,19 @@
...
+
+        <ol>
+            <li>The Reading System provides the user with a way to turn text reading on</li>
+            <li>If turned on, the extract on the next page is read aloud by the Reading System</li>
+            <li>If the reading system highlights the active text, then the highlight stops at the second section</li>
+            <li>The full text is read</li>
```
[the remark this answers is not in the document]
iherman: Not exactly. In the case of the exceeding clip-end the audio jumps to the next audio segment mentioned in the SMIL file. In this case the this 'portion' is the last in the SMIL file, meaning that reading will continue for the full length of the audio track.

tests/mol-audio_no_clipbegin/OPS/mo/mobydick.smil
```diff
@@ -0,0 +1,20 @@
...
+    <body>
+        <seq epub:textref="../mobydick.xhtml#mobyexcerpt">
+            <par id="first">
+                <text src="../mobydick.xhtml#first"/>
+                <audio src="../audio/mobydick.mp3" clipEnd="0:00:44.783"/>
+            </par>
+
+            <par id="second">
+                <text src="../mobydick.xhtml#second"/>
+                <audio src="../audio/mobydick.mp3" clipBegin="0:00:44.783" clipEnd="0:00:50.450"/>
```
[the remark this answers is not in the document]
iherman: I do not think so. These are mutually independent features and requirement, hence the separate tests.

tests/mol-css_classes/OPS/mobydick.xhtml
iherman: +1

tests/mol-timing-synchronization_multiple_audio/EPUB/package.opf
```diff
@@ -12,7 +12,8 @@
     <link rel="dcterms:rights" href="https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document"/>
     <link rel="dcterms:rightsHolder" href="https://www.w3.org"/>
     <meta property="belongs-to-collection">should</meta>
-    <meta property="dcterms:isReferencedBy">https://www.w3.org/TR/epub-rs-33/#mol-audio-exceeding-clipend</meta>
+    <meta property="dcterms:isReferencedBy">https://www.w3.org/TR/epub-rs-33/#mol-timing-sync</meta>
+    <meta property="dcterms:isReferencedBy">https://www.w3.org/TR/epub-rs-33/#mol-rendering-audio</meta>
     <meta property="dcterms:modified">2022-02-14T00:00:00Z</meta>
     <meta property="media:duration" refines="#md-smil">00:02:47.768</meta>
```
marisademeglio, line 18: I believe this duration value is wrong. 
One day, I will get around to writing a utility that fills this in automatically, although by then, we will probably be done writing MO tests.

marisademeglio (changes requested):
Needs duration edit (see inline comment)

tests/mol-timing-synchronization_multiple_audio/EPUB/package.opf
iherman: Sigh... yes, indeed. I think the right value is `00:01:46.35`. I use a [time calculator](http://www.unitarium.com/time-calculator) which, hopefully, is correct.

We are violating the DRY principle here, b.t.w. Wouldn't it be enough to require the duration per smil files and the overall media duration would then be the sum of all those? But anyway, that is a discussion for another day...

Thx!

PR #122, opened 2022-02-16
MO Tests on text-to-speech fallback

### LINKED ISSUE #121: TTS fallback of Media Overlay
Will prepare a test for the [Text-to-Speech](https://w3c.github.io/epub-specs/epub33/rs/#sec-text-to-speech) section.

iherman:
Two tests: the smil file do not refer to any audio file at all, only to text portions. The 'single' means there is only one reference in the smil file, the 'multiple' means there are several references to different paragraphs in the same text.

Expected to hear the full text read aloud by the Reading System's text-to-speech engine

Fix #121


iherman:
@marisademeglio @mattgarrish I would really need your review on this to see if I interpreted the spec properly. None of the engines I have on my machine implements the expected output :-(

marisademeglio:
I think that your test is fine. I have never seen this implemented in the wild though for some reason I thought that Thorium supported it. @danielweck do you know if that's true?

iherman:
> I think that your test is fine. I have never seen this implemented in the wild though for some reason I thought that Thorium supported it. @danielweck do you know if that's true?

Well... it is not working in Thorium either... (but maybe I have to do some special setting to get TTS working?)

danielweck:
hello, this feature was implemented in Readium "1", but to my knowledge it is not supported in the second generation Readium toolkit(s) (desktop and mobile SDKs)

iherman:
> I think that your test is fine. I have never seen this implemented in the wild though for some reason I thought that Thorium supported it. @danielweck do you know if that's true?

Well, that is what the tests are for. If we do not get enough successful implementations, then we will have to mark it somehow in the spec. Note that the whole of MO is a _SHOULD_, so it is not in the critical path of publishing a Rec.

@dlazin can you have a look at these? They are simple :-)

tests/mol-tts_multi/EPUB/content_001.xhtml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<html xmlns="http://www.w3.org/1999/xhtml">
+<head>
+    <meta charset="utf-8" />
+    <title>Multiple text-to-speech fallback for MO</title>
+</head>
+<body>
+    <p>Test passes (i.e., the Reading System correctly falls back on Text-to-Speech when no audio file is present) if</p>
```
dlazin, line 8:
```suggestion
    <p>Test passes (i.e., the Reading System correctly falls back to its own Text-to-Speech system when no audio file is present) if</p>
```

dlazin: (This is just for clarity, since I didn't understand on first reading)

tests/mol-tts_multi/EPUB/content_001.xhtml
```diff
@@ -0,0 +1,15 @@
...
+    <ol>
+        <li>The Reading System provides the user with a way to turn text reading on</li>
+        <li>If turned on, the extract on the next page is read aloud by the Reading System using Text-to-Speech</li>
+        <li>If the Reading System implements the association with style information, the text being read is highlighted with a green background and the rest of the text is grayed out</li>
```
dlazin, line 12: Does the "if" here imply that the test can still pass even if this last case fails, because the RS doesn't support CSS? If so I think this should be a note or something, not part of the list, and should be clearly marked as not required to pass.

tests/mol-tts_multi/EPUB/package.opf
```diff
@@ -0,0 +1,31 @@
...
+    <dc:creator>Ivan Herman</dc:creator>
+    <dc:date>2022-02-17</dc:date>
+    <dc:description>Tests whether the Reading System implements Text-to-Speech as a fallback when no audio content is provided. Severak fragments are defined in the SMIL file.</dc:description>
+    <dc:identifier id="pub-id">mol-tts_multi</dc:identifier>
+    <dc:language>en</dc:language>
+    <dc:title>Multiple text-to-speech fallback for MO</dc:title>
+    <link rel="dcterms:rights" href="https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document"/>
+    <link rel="dcterms:rightsHolder" href="https://www.w3.org"/>
+    <meta property="belongs-to-collection">should</meta>
+    <meta property="dcterms:isReferencedBy">https://www.w3.org/TR/epub-rs-33/#sec-text-to-speech</meta>
```
dlazin, line 14: Since there are two separate normative statements in this section, I think they should have distinct anchors and you should point to the one under test (or both individually).

tests/mol-tts_multi/EPUB/package.opf
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<package xmlns="http://www.idpf.org/2007/opf" xmlns:epub="http://www.idpf.org/2007/ops" version="3.0" xml:lang="en" unique-identifier="pub-id">
+  <metadata xmlns:dc="http://purl.org/dc/elements/1.1/">
+    <dc:coverage>Media Overlays</dc:coverage>
+    <dc:creator>Ivan Herman</dc:creator>
+    <dc:date>2022-02-17</dc:date>
+    <dc:description>Tests whether the Reading System implements Text-to-Speech as a fallback when no audio content is provided. Severak fragments are defined in the SMIL file.</dc:description>
+    <dc:identifier id="pub-id">mol-tts_multi</dc:identifier>
+    <dc:language>en</dc:language>
+    <dc:title>Multiple text-to-speech fallback for MO</dc:title>
```
dlazin, line 10: Nit: I recommend avoiding abbreviations like MO (outside of the test IDs, that is), and also I think "for MO" is unnecessary in the title. Remove throughout?

Consider: Text-to-speech fallback for multiple fragments

tests/mol-tts_single/EPUB/package.opf
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<package xmlns="http://www.idpf.org/2007/opf" xmlns:epub="http://www.idpf.org/2007/ops" version="3.0" xml:lang="en" unique-identifier="pub-id">
+  <metadata xmlns:dc="http://purl.org/dc/elements/1.1/">
+    <dc:coverage>Media Overlays</dc:coverage>
+    <dc:creator>Ivan Herman</dc:creator>
+    <dc:date>2022-02-17</dc:date>
+    <dc:description>Tests whether the Reading System implements Text-to-Speech as a fallback when no audio content is provided. Only one fragment is defined in the SMIL file.</dc:description>
+    <dc:identifier id="pub-id">mol-tts_single</dc:identifier>
+    <dc:language>en</dc:language>
+    <dc:title>Single text-to-speech fallback for MO</dc:title>
```
dlazin, line 10: "Text-to-speech fallback for a single fragment" or "Single text-to-speech fallback"? (no "for MO")

tests/mol-tts_multi/EPUB/package.opf
```diff
@@ -0,0 +1,31 @@
...
+    <link rel="dcterms:rights" href="https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document"/>
+    <link rel="dcterms:rightsHolder" href="https://www.w3.org"/>
+    <meta property="belongs-to-collection">should</meta>
+    <meta property="dcterms:isReferencedBy">https://www.w3.org/TR/epub-rs-33/#sec-text-to-speech</meta>
+    <meta property="dcterms:modified">2022-02-17T00:00:00Z</meta>
+    <meta property="media:duration" refines="#md-smil">00:01:46.35</meta>
+    <meta property="media:duration">00:01:46.35</meta>
```
dlazin, line 17: Are these two lines right? The spec says "The implicit duration of the text element (and by inference, of the parent par element) is therefore determined by the execution of the Text-to-Speech engine, and cannot be known at authoring time (factors like speech rate, pauses and other prosody parameters influence the audio output)."

tests/mol-tts_single/EPUB/content_001.xhtml
```diff
@@ -0,0 +1,15 @@
...
+<head>
+    <meta charset="utf-8" />
+    <title>Single text-to-speech fallback for MO</title>
+</head>
+<body>
+    <p>Test passes (i.e., the Reading System correctly falls back on Text-to-Speech when no audio file is present) if</p>
+    <ol>
+        <li>The Reading System provides the user with a way to turn text reading on</li>
+        <li>If turned on, the extract on the next page is read aloud by the Reading System using Text-to-Speech</li>
+        <li>If the Reading System implements the association with style information, the text being read is highlighted with a green background and the rest of the text is grayed out</li>
```
dlazin, line 12: Same question about whether this is a true requirement. Either the "if" should change, or this should be clearly marked as optional.

tests/mol-tts_single/EPUB/package.opf
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<package xmlns="http://www.idpf.org/2007/opf" xmlns:epub="http://www.idpf.org/2007/ops" version="3.0" xml:lang="en" unique-identifier="pub-id">
+  <metadata xmlns:dc="http://purl.org/dc/elements/1.1/">
+    <dc:coverage>Media Overlays</dc:coverage>
+    <dc:creator>Ivan Herman</dc:creator>
+    <dc:date>2022-02-17</dc:date>
+    <dc:description>Tests whether the Reading System implements Text-to-Speech as a fallback when no audio content is provided. Only one fragment is defined in the SMIL file.</dc:description>
```
dlazin, line 7: Nit: we never capitalize Reading System in the descriptions (see https://w3c.github.io/epub-tests/). I don't really care about the content of the test file, but let's try to keep this consistent in test descriptions so the test document is clean.

tests/mol-tts_multi/EPUB/package.opf
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<package xmlns="http://www.idpf.org/2007/opf" xmlns:epub="http://www.idpf.org/2007/ops" version="3.0" xml:lang="en" unique-identifier="pub-id">
+  <metadata xmlns:dc="http://purl.org/dc/elements/1.1/">
+    <dc:coverage>Media Overlays</dc:coverage>
+    <dc:creator>Ivan Herman</dc:creator>
+    <dc:date>2022-02-17</dc:date>
+    <dc:description>Tests whether the Reading System implements Text-to-Speech as a fallback when no audio content is provided. Severak fragments are defined in the SMIL file.</dc:description>
```
dlazin, line 7:
```suggestion
    <dc:description>Tests whether the reading system implements Text-to-Speech as a fallback when no audio content is provided. Several fragments are defined in the SMIL file.</dc:description>
```

tests/mol-tts_single/EPUB/package.opf
```diff
@@ -0,0 +1,31 @@
...
+    <dc:identifier id="pub-id">mol-tts_single</dc:identifier>
+    <dc:language>en</dc:language>
+    <dc:title>Single text-to-speech fallback for MO</dc:title>
+    <link rel="dcterms:rights" href="https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document"/>
+    <link rel="dcterms:rightsHolder" href="https://www.w3.org"/>
+    <meta property="belongs-to-collection">should</meta>
+    <meta property="dcterms:isReferencedBy">https://www.w3.org/TR/epub-rs-33/#sec-text-to-speech</meta>
```
dlazin, line 14: Same comment — I think this should be one or two more specific anchors, since there are two normative statements here.

tests/mol-tts_single/EPUB/package.opf
```diff
@@ -0,0 +1,31 @@
...
+    <meta property="dcterms:modified">2022-02-17T00:00:00Z</meta>
+    <meta property="media:duration" refines="#md-smil">00:01:46.35</meta>
+    <meta property="media:duration">00:01:46.35</meta>
```
dlazin, line 17: Same comment — is duration a mistake here?

dlazin (approved):
Apologies again for such a slow review — too busy! But I will keep catching up like this, even if not promptly.

Also EXCELLENT TEST IDs :)

tests/mol-tts_multi/EPUB/content_001.xhtml
dlazin: Also, the language choice here is a bit weird (stilted?). Can you say something like "If the Reading System implements styling for media overlays, ..."

tests/mol-tts_single/EPUB/content_001.xhtml
dlazin: And again, maybe something like "If the Reading System implements styling for media overlays, ..."

tests/mol-tts_multi/EPUB/content_001.xhtml
iherman: > Does the "if" here imply that the test can still pass even if this last case fails, because the RS doesn't support CSS?

Well... it is a gray area, in fact (the same terminology is there for a number of other tests). Yes, support for CSS is optional; however *if* the RS implements it, then the described behaviour is the required one. Ie, the test fails if it does not really do it the way it described.

For the sake of consistency with around 4-5 other tests, let us leave it as is.

tests/mol-tts_multi/EPUB/package.opf
iherman: Done.

tests/mol-tts_multi/EPUB/package.opf
iherman: Right. I have not yet handled the anchors; doing it this time.

tests/mol-tts_multi/EPUB/package.opf
iherman: Good catch. And I think the spec is a bit unclear on this, so I will have to raise an issue for clarification. The fact is: per https://w3c.github.io/epub-specs/epub33/core/#sec-duration the presence of a `media:duration` is _required_ per spec, and epubcheck complains if it is not present (as it should, per spec). This also means that, in the case of TTS, those values are arbitrary. Which is still all right, but I believe the spec text in https://w3c.github.io/epub-specs/epub33/rs/#sec-text-to-speech.

For the moment, I leave these in the OPF file.

tests/mol-tts_single/EPUB/content_001.xhtml
iherman: Same answer :-)

tests/mol-tts_single/EPUB/package.opf
iherman: +1

tests/mol-tts_single/EPUB/package.opf
iherman: +1 (took the former)

tests/mol-tts_single/EPUB/package.opf
iherman: +1, like above

tests/mol-tts_single/EPUB/package.opf
iherman: Same answer :-)

PR #131, opened 2022-03-04
Adding ocf-url tests

iherman:
Adding simple tests that check the handling of relative URL-s. All three refer to an image to be displayed via an `img` tag.

Note that two out of the three tests _are not valid_ in terms of epubcheck, but they should still be properly handled by the Reading System per the defined algorithm.

---

 @dlazin, do we want to think about another metadata tag marking the test as 'invalid EPUB, but that is all right'?

PR #133, opened 2022-03-04
Added two url parsing tests for the OCF URL section

iherman:
This may be controversial, unnecessary, or plainly wrong...

I tried to create two tests as an equivalent to #131, which shows the result of URL parsing by the RS in a script. It may be useful for the tester if the tests in #131 go wrong (some sort of a debug statement).

The problem is that the display of the container Root URL is, by definition iffy, and I am not sure what is there holds. I have not found a better way.

We may decide to remove these tests altogether (that is why they are in a separate PR!)


tests/ocf-url_parse_leaking_relative/EPUB/content_001.xhtml
```diff
@@ -0,0 +1,60 @@
...
+                    document.getElementById('expected').textContent = `{Container Root URL}${expected_path}`;
+                } catch(e) {
+                    alert(e)
+                }
+            }
+            window.addEventListener('load', test_url);
+        </script>
+    </head>
+    <body>
+        <p id="no_scripting" style="display: block">This reading system does not support scripting, which is necessary for this test to run.</p>
```
rdeltour, line 42: it's a detail and not specific to the test itself: but this could probably rely on HTML's `noscript` element?

tests/ocf-url_parse_leaking_relative/EPUB/content_001.xhtml
```diff
@@ -0,0 +1,60 @@
...
+                    // 'pure' URL. This is done by cloning the document's URI and removing  
+                    // relevant portions of the full URL
+                    // (There are RS-s that add a long query string to the URL, hence the necessity of this step)
+                    const new_URL = new URL(document.documentURI);
+                    new_URL.search = '';  new_URL.hash = '';
+                    const documentURI = new_URL.toString();
+
+                    // This is, hopefully, a good representation of the container root URL: remove the
+                    // relative path of the current document
+                    const container_root_URL = documentURI.slice(0,documentURI.length - current_path.length);
```
rdeltour, line 25: couldn't all this be replaced by:

```js
const container_root_URL = `new URL('../..', document.documentURI)
```

tests/ocf-url_parse_leaking_relative/EPUB/content_001.xhtml
iherman: When I first read the comment, my reaction was "Oops...". But then I got suspicious and found this:

> The [noscript](https://html.spec.whatwg.org/multipage/scripting.html#the-noscript-element) element is only effective in [the HTML syntax](https://html.spec.whatwg.org/multipage/syntax.html#syntax), it has no effect in [the XML syntax](https://html.spec.whatwg.org/multipage/xhtml.html#the-xhtml-syntax). This is because the way it works is by essentially "turning off" the parser when scripts are enabled, so that the contents of the element are treated as pure text and not as real elements. XML does not define a mechanism by which to do this.

See [HTML spec](https://html.spec.whatwg.org/multipage/scripting.html#the-noscript-element). I guess that means we should not use it in EPUB...

tests/ocf-url_parse_leaking_relative/EPUB/content_001.xhtml
iherman: What made me stop and do what I did was if the RS uses a URL for the document of the form:

```
file://path/to.epub#path=/EPUB/content.opf
```
or

```
jar:file:/path/to.epub!/EPUB/content.opf
```

(these are [your examples](https://github.com/w3c/epub-specs/issues/1374#issuecomment-847978623)😉)

But, on the other hand, if the parsing of `../..` go wrong, that means that RS is non-conform anyway... which is a problem with the RS then...

(And this may make the previous clean-up of the hash and search also unnecessary.)

tests/ocf-url_parse_leaking_relative/EPUB/content_001.xhtml
rdeltour: Oh, TIL. Good catch!

tests/ocf-url_parse-path-absolute/EPUB/content_001.xhtml
```diff
@@ -0,0 +1,49 @@
+<!DOCTYPE html>
+<html xmlns="http://www.w3.org/1999/xhtml" lang="en-US" xml:lang="en-US">
+    <head>
+        <meta charset="utf-8"/>
+        <title>Parse a root-relative URL</title>
```
dlazin, line 5: "Root-relative image reference, parsed by JS"?

Or something better than that; I'm not clear on the distinction between this and https://github.com/w3c/epub-tests/blob/main/tests/ocf-url_link-path-absolute/EPUB/package.opf, other than the JS. I think their titles should be as close as possible.

tests/ocf-url_parse-leaking-relative/EPUB/content_001.xhtml
```diff
@@ -0,0 +1,49 @@
...
+
+                    const container_root_URL = new URL('../..', document.documentURI);
+                    document.getElementById('container_root_url').textContent = container_root_URL;
+
+                    document.getElementById('test_url').textContent = parent_relative;
+
+                    const image_src = new URL(parent_relative, container_root_URL);
+                    document.getElementById('image_url_parent_paths').textContent = image_src;
+
+                    document.getElementById('expected').textContent = `{Container Root URL}${expected_path}`;
```
dlazin, line 22: I'm confused: what's {Container Root URL}? I would expect that to be a JS var, but I don't see "Container Root URL" defined (and, of course, it has spaces). Does this somehow magically interpret container_root_URL?

My other comment is that I think this test would be more readable if you defined "expected" as a variable at the top of the file, but I can't tell whether "expected" genuinely depends on stuff inside this function.

tests/ocf-url_parse-path-absolute/EPUB/package.opf
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<package xmlns="http://www.idpf.org/2007/opf" xmlns:epub="http://www.idpf.org/2007/ops" version="3.0" xml:lang="en" unique-identifier="pub-id">
+  <metadata xmlns:dc="http://purl.org/dc/elements/1.1/">
+    <dc:coverage>Open Container Format</dc:coverage>
+    <dc:creator>Ivan Herman</dc:creator>
+    <dc:date>2021-03-05</dc:date>
+    <dc:description>Parse a path-absolute URL, i.e., starting with a slash, from the content to an image. The folder hierarchy containing the image starts at the root level.</dc:description>
```
dlazin, line 7: "beginning with a leading slash" for consistency with https://github.com/w3c/epub-tests/blob/main/tests/ocf-url_link-path-absolute/EPUB/package.opf

dlazin (approved):
Can you give some more context? When https://www.w3.org/TR/epub-33/#sec-container-iri-step-parse refers to "parsing," it doesn't mean JS specifically, does it? Obviously JS parsing *could* differ from the XHTML parsing, so they're both worth testing, but this test seems to imply to me that "parsing" is somehow distinct from the tests you've already submitted.

If, as I presume, these both count as parsing, maybe we should clarify the other tests to say "parsed from XHTML" or something, by comparison to these?

tests/ocf-url_parse-leaking-relative/EPUB/content_001.xhtml
iherman: I have added comments above

iherman: Also, the `{Container Root URL}` is to be displayed literally on the page. The problem is that the value of this thing is, sometimes, unreadably long (e.g., in Thorium) so displaying the value here makes the test unmanagable. It refers to the value with the same label a few paragraphs down.

tests/ocf-url_parse-path-absolute/EPUB/package.opf
iherman: yep

tests/ocf-url_parse-path-absolute/EPUB/content_001.xhtml
iherman: done

PR #136, opened 2022-04-30
Allowing the 'null' value for implementation report.

iherman:
This is to fulfill the request of @bduga: implementation reports allow for the value of `null`, resulting in a `n/a` cell in the report. This is also the default value in the implementation report template.

The PR also includes some additional minor changes:

- removal of the test report created during the live demonstration
- minor changes on the respec headers to avoid unnecessary respec warnings and inconsistencies in the generated header


iherman:
@dlazin I would like to merge this PR as soon as possible to avoid unnecessary merge conflicts if new tests come in...

docs/drafts/index.html
```diff
@@ -7,13 +7,13 @@
...
 				specStatus: "base",
 				noRecTrack: true,
-				publishDate: "2022-04-22",
+				publishDate: "2022-04-30",
 				shortName: "epub-tests",
-				edDraftURI: "https://w3c.github.io/epub-tests/results/",
+				edDraftURI: "https://w3c.github.io/epub-tests/results/drafts/",
                 copyrightStart: "2021",
 				editors:[
 					{
```
dlazin, line 19: Unrelated to this PR, but reminder that we were thinking of removing http://www.aptest.com/standards/htmldiff/htmldiff.pl (line 38) from all respec configs.

src/lib/html.ts
```diff
@@ -101,7 +101,12 @@ function create_impl_reports(data: ReportData): string {
             add_child(header_row, 'th', 'Id');
             add_child(header_row, 'th', 'Req');
             for (const impl of (consolidated ? data.consolidated_implementers : data.implementers)) {
-                const head = impl.variant !== 'consolidated' ? `${impl.name}<br />${impl.variant}` : impl.name;
+                let head: string;
+                if (impl.variant !== undefined) {
+                    head = impl.variant !== 'consolidated' ? `${impl.name}<br />${impl.variant}` : impl.name;
```
dlazin, line 106: I'm having trouble parsing this (especially telling how it's different from the old version) — could use a comment.

dlazin (approved):
Sorry for the slow review, and thanks for doing this!

src/lib/html.ts
iherman: The variant field in the implementation report is optional, and the original code did not take that into account. As a result, in some cases, an extra "(undefined)" field appeared in the output. Bug...

docs/drafts/index.html
iherman: You are right. I have removed these before merging.

PR #138, opened 2022-05-12
Changed the way the should/may tested are presented.

### LINKED ISSUE #101: Color Contrast issue with "should"'s
In the current table view of all the issues the Must issues all look fine, but the "Should" issues seem like they are grayed out which makes it hard to read and fails WCAG 2.4.3 for normal text size fonts.

Since we have the column for "Req" containing Must/Should etc. we don't really have to also visually gray out these rows.
Now perhaps just the words "Must" "Should" etc could had a different background

Black on Red - Must
Black on Magenta - Should
Black on Yellow - May
as shown here:
Screen Shot: Text Must/Should/May with color backgrounds

iherman:
This is a response to the long standing issue #101 raised by @clapierre, but taking a very different approach. The changes to show (or not) tests for SHOULD/MAY statements are as follows:

- by default, the rows referring to SHOULD /MAY tests look exactly the same as the others (except for the relevant table cell that indicates the MUST/SHOULD/MAY 'status' of the test
- I have added a button in the introductory part that makes it possible for the user to switch between the visibility/hidden status of the rows. Ie, the user may choose to ignore the SHOULD/MAY tests to concentrate on the MUST tests only.

Note that the requested visibility status value stays on for the user on that browser and is carried over the two files (test description and implementation report). Technically, the value is stored in the browser's local storage.

Fix #101

Here is a [preview on githack](https://raw.githack.com/w3c/epub-tests/script/optional-display/docs/index.html)

--- 

I also made some cosmetic changes: the CSS and the (new) JS files are now embedded in the respec sources; this avoid the awkward situation of having those files twice on the repository.


docs/drafts/index.html
```diff
@@ -75,6 +205,12 @@ <h2>Introduction</h2>
 				<li>References to the corresponding normative statements or sections in the EPUB 3.3 specification(s).</li>
 				<li>Pointer to the (consolidated) implementation status of the test in the separate <a href="results">implementation report</a>.</li>
 			</ol>
+
+			<p>
+				Strictly speaking, the “<code>should</code>” and “<code>may</code>” tests are not necessary for the official CR testing of
```
dlazin, line 210: I don't think you need both quotation marks and <code>; the <code> acts as implicit punctuation.

docs/drafts/index.html
```diff
@@ -75,6 +205,12 @@ <h2>Introduction</h2>
...
+				the specifications. These tests are currently <span id="test_visibility"></span> in the tables below; by pushing the 
+				<button id="should_may_visibility">switch visibility</button> button the visibility of these tests may turned on/off.
```
dlazin, line 212: I would say "to change the visibility, click the switch visibility button."

dlazin (approved):

dlazin:
> Works great Ivan, but seems kinda buried above, I think its fine to discuss that button where you have it but I would like to see down in Section 2 the actual button which would be "Show only MUST tests" and when you have switched it would toggle to "Show all Tests"

Yes, I agree.

docs/drafts/index.html
iherman: +1

docs/drafts/index.html
iherman: +1

iherman:
> Works great Ivan, but seems kinda buried above, I think its fine to discuss that button where you have it but I would like to see down in Section 2 the actual button which would be "Show only MUST tests" and when you have switched it would toggle to "Show all Tests"

Yep, that is a great idea, @clapierre, thanks.

I have implemented it, although the button's label is a bit more verbose (but that should be all right).

I will merge this PR now; the reason is that it involves merging files that are also re-generated by the github actions when tests are updated, and this can lead to problematic merge conflicts... But I would not consider this as closed insofar as I am happy to make updates if you have other suggestions!

PR #140, opened 2022-05-12
Update page-layout-both-spread test

wareid:
Fixing test anchor, but also adding second part of the test.


dlazin (approved):
Add the new page to nav.xhtml, please!

dlazin (approved):
Thanks!

PR #145, opened 2022-05-20
Remove mention of online EPUB Validator

dlazin:


iherman (approved):
Just a note to ourselves: [epubcheck 5](https://github.com/w3c/epubcheck/releases) is in beta, and I think it will be available more widely soon (@rdeltour, I hope that is correct). When this is true, it may be worth emphasizing to use that version instead of version 4, because it follows the requirements and syntax of EPUB 3.3.

PR #148, opened 2022-05-24
Standardize formatting of mol-navigation, add EPUB, and delete mol-audio-rendering

dlazin:
This addresses most of my comments from https://github.com/w3c/epub-tests/pull/117 (as well as a bunch of basic formatting that I didn't ask for because it's too nitpicky, but I don't mind doing it myself). This deletes mol-rendering-audio because, in that discussion, (I think) we realized it's redundant with more granular tests that Ivan wrote.


tests/mol-navigation/EPUB/package.opf
```diff
@@ -1,41 +1,37 @@
...
-        <dc:coverage>Media Overlays</dc:coverage>
-        <dc:creator>Marisa DeMeglio</dc:creator>
-        <dc:date>2022-02-10</dc:date>
-        <dc:description>When navigating to another page, the media overlay playback position updates accordingly.</dc:description>
-        <dc:identifier id="pub-id">mol-navigation</dc:identifier>
-        <dc:language>en</dc:language>
-        <dc:title>Navigation with media overlays</dc:title>
-        <link rel="dcterms:rights" href="https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document"/>
-        <link rel="dcterms:rightsHolder" href="https://www.w3.org"/>
-        <meta property="belongs-to-collection">should</meta>
```
dlazin, line 14: Note: I removed this entirely, because https://www.w3.org/TR/epub-rs-33/#sec-rsconf-navigation appears to be a MUST, not a SHOULD. Please let me know if I misunderstood something.

tests/mol-rendering-audio/EPUB/package.opf
```diff
@@ -4,33 +4,31 @@
...
+    <dc:description>
+      The reading system must respect the requirements for clipBegin and clipEnd when playing audio.
+    </dc:description>
     <dc:identifier id="pub-id">mol-rendering-audio</dc:identifier>
     <dc:language>en</dc:language>
-    <dc:title>Media Overlays Tests: Rendering Audio</dc:title>
+    <dc:title>Media overlays render audio</dc:title>
     <link rel="dcterms:rights" href="https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document"/>
     <link rel="dcterms:rightsHolder" href="https://www.w3.org"/>
-    <meta property="belongs-to-collection">should</meta>
```
dlazin, line 13: Same: I removed this entirely, because https://www.w3.org/TR/epub-rs-33/#sec-rsconf-rendering-audio appears to be a MUST, not a SHOULD (although it has one SHOULD at the end, which I think is untested). Please let me know if I misunderstood something.

tests/mol-navigation/EPUB/package.opf
dlazin: Oh, wait: @iherman are all MO tests marked as SHOULD because the overall section is a SHOULD, even though these requirements are a MUST within the SHOULD?

I'm not sure what makes sense here:
a) From a CR perspective, we need to make sure all the MUSTs pass in at least 2 RSes, even if controlled by a SHOULD, to prove that the spec is implementable. This implies that all MO MUSTs should not be labeled as SHOULD with this attribute.

b) From an RS perspective, your system is compliant if it implements all the top-level MUSTs, and it may safely ignore this test.

IMO a) is more critical for us, and thus MO tests should not be marked as SHOULD. Thoughts?

iherman (approved):

tests/mol-navigation/EPUB/package.opf
iherman: > Oh, wait: @iherman are all MO tests marked as SHOULD because the overall section is a SHOULD, even though these requirements are a MUST within the SHOULD?

yes, that was the reason.

> IMO a) is more critical for us, and thus MO tests should not be marked as SHOULD. Thoughts?

Yes, I guess that is indeed a better choice. A bit more dangerous, though: I am almost sure that some of the MUST statements will not have been implemented but, oh well, that is what testing is all about (and we will mark them as under-implemented).

Thanks for provoking these thoughts! 😀 But I would like to get at least @wareid or @dauwhe chime in on these.

tests/mol-rendering-audio/EPUB/package.opf
iherman: Same answer...

tests/mol-navigation/EPUB/package.opf
dlazin: Restored for now (for consistency). Let's discuss in the next meeting, and if we decide these should not be SHOULDs, then we can pull them all with one PR.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2022-05-26-epub#section2-1) on 2022-05-26

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 2.1. Standardize formatting of mol-navigation.

_See github pull request [epub-tests#148](https://github.com/w3c/epub-tests/pull/148)._




**Dan Lazin:** this is a discussion of a comment in the PR.  
… this was a test for MO, and this was a test for a _MUST_ within the MO section (which is a _SHOULD_).  
… Ivan has created a way to filter should tests, so that, for example, a tester could say "i'm not worried about the 'should' tests".  
… but in order for us to determine whether the spec is implementable, the _MUST_s within the should sections are still important.  
… we need to have 2 independent implementations of these _MUST-s_ too.  
… so what is the purpose of marking some tests as should, and based on the answer to same, how do we treat _MUST-s_ in _SHOULD_ sections?.  

**Dave Cramer:** do we have this problem on a higher level? If we allow possibility of audio only RS, then anything related to visual only is not required.  

**Dan Lazin:** that's the opposite of this, because those tests would not be marked as _SHOULD-s_, and could not be accidentally filtered out.  

**Dave Cramer:** so in some sense we may need to leave it up to the RS dev, e.g. if the RS in question doesn't support MO.  

**Matt Garrish:** so we're concerned about how to handle filtering? Can Ivan just fix this in the filtering script?.  

**Dan Lazin:** I think its what is the purpose of marking a test as a should test? Are we telling RS tester not to worry about this? Or are we telling Director not to worry about this?.  
… I think purpose of our tests is CR. So I would think we do NOT mark such tests as should tests.  
… maybe we should defer this until Ivan is here, this isn't urgent.  

**Brady Duga:** yes, fine to defer. But it seems these tests are going to be marked as _MUST-s_ because they ARE _MUST-s_..  
… otherwise they would just be _SHOULD_ in the spec.  

**Wendy Reid:** I would be more worried about this if it wasn't in a feature like MO that is already fairly well supported.  
… could some of the more technical features of MO be at risk? Sure. But the core feature of MO I'm not worried about.  

**Dan Lazin:** hearing duga's reply, and no disagreements, I'm tending towards not flagging them as should tests. I don't think Ivan feels strongly, but we can follow up with him next meeting.  

**Wendy Reid:** I'll wait for the PR for us to resolve.  

</details>

iherman:
For the records, because it was mentioned in the minutes of the last meeting: I would have voted +1 had I been on the call.

PR #150, opened 2022-05-25
Adding missing i18n tests

### LINKED ISSUE #147: Missing I18n tests
I will take care of the missing i18n tests (ie, rows 2-5 of the [missing tests' sheet](https://docs.google.com/spreadsheets/d/13wgptApSVbQbYzi2L5VXe3D_1QLzJOTGySUAtOxn4sA/edit#gid=0)).

iherman:
These tests correspond to the statements in [§2.6 Internationalization](https://w3c.github.io/epub-specs/epub33/rs/#sec-epub-rs-i18n) that still need tests. Two remarks:

- epubcheck fails on the `cnt-i18n_svg_direction` test. See https://github.com/w3c/epubcheck/issues/1334 and https://github.com/w3c/svgwg/issues/884; TL;DR: the `direction` attribute does not seem to be properly specified in the SVG spec (but it works on browsers). We can expect epubcheck to take care of this.
- There is (at the moment) no test for the second half of the statement on the usage of `lang` in SVG file. I have not found a way to visually differentiate an SVG file in function of the attribute value. (The only thing I though of is to use a script to inquire the language value of an element and report that, but that sounds like a sledgehammer.) Not 100% what to do about this.

Fix #147


iherman:
Pinging @dlazin to review these set of tests...

Noting, however, that all the _MUST_ statements addressed by these tests may be seen as irrelevant as far as the RS spec is concerned. Indeed, the fact that the RS must honour the `lang` and `dir` attributes for XHTML and SVG can be seen as part of the general statement  in [§2](https://www.w3.org/TR/epub-rs-33/#sec-pub-resources) whereby:

> Reading systems MUST process [publication resources](https://www.w3.org/TR/epub-33/#sec-publication-resources) [[epub-33](https://www.w3.org/TR/epub-rs-33/#bib-epub-33)].

Processing "normatively" XHTML and SVG does mean to process these attributes, whatever that means in HTML and SVG. Ie, it goes without saying that a conformant RS would do that and, consequently, the statements we are talking about in [2.6](https://www.w3.org/TR/epub-rs-33/#sec-epub-rs-i18n) may be considered as superfluous; after all, we do not explicitly say, nor do we test, all the zillion of other statements related to, e.g., XHTML processing.

I do _**not**_ propose to remove those statements from the spec (they are not causing any harm); removing those could be misunderstood and may become a "political" problem. But the tests in this section would then be enough for that section in my view, in spite of the problem with testsing `lang`+SVG.

mattgarrish:
> It does not say anything whether you have inherited the attribute from somewhere up in the DOM tree.

I'm not sure I understand what you're saying here. documentElement is the root element, so there's nowhere else the values could have been inherited from. If you don't get the value that was set, or a null value, then the reading system injected the values.

> A script should then reproduce the inheritance rule but then the only information would be that the attribute is _somewhere_.

This is why it's pointless to try and test if SVG user agents respect the languages that have been set. It's outside the scope of what we even control.

> But that is not what the spec says right now...

Right, but my point above is that what the spec says is not our problem to enforce. Why are we saying anything about how language and direction are required to be processed in any format? The only epub-specific part is that the reading system not inject information in the absence of it.

iherman:
> Right, but my point above is that what the spec says is not our problem to enforce. Why are we saying anything about how language and direction are required to be processed in any format? The only epub-specific part is that the reading system not inject information in the absence of it.

So the minimal step we can do is... to remove [§3.6](https://www.w3.org/TR/epub-rs-33/#sec-epub-rs-i18n) altogether! Normatively speaking, nothing would change if we do this.

mattgarrish:
> Normatively speaking, nothing would change if we do this.

It would be valid for reading systems to set/override the language and direction of a document, but I think we've taken this to extremes anyway. I don't think this was ever the problem.

Section 3.6 appears to have been built out from this lonely little statement in 3.2 for dc:language:

> This value is not inherited by the individual resources of the Rendition.

That's all we said prior to this revision, and that statement was only to emphasize to authors that dc:language does not stand in as a replacement for xml:lang in the package document or any method in a content document. We even still have it as a note in dc:language:

> Publication resources do not inherit their language from the dc:language element(s). EPUB creators must set the language of a resource using the intrinsic methods of the format.

The solution has become bigger than the problem, given that accessibility checking will flag missing languages in package/content documents anyway.

iherman:
> Section 3.6 appears to have been built out from this lonely little statement in 3.2 for dc:language:
> 
> > This value is not inherited by the individual resources of the Rendition.
> 
> That's all we said prior to this revision, and that statement was only to emphasize to authors that dc:language does not stand in as a replacement for xml:lang in the package document or any method in a content document. We even still have it as a note in dc:language:
> 
> > Publication resources do not inherit their language from the dc:language element(s). EPUB creators must set the language of a resource using the intrinsic methods of the format.

And we even have tests for those 😀

It is obviously a WG's decision, but I would be in favour of simply nuking that section (making the tests I made obsolete... Oh well).

I can prepare a PR on this later today.

tests/cnt-i18n_svg-direction/EPUB/content_001.xhtml
```diff
@@ -0,0 +1,9 @@
+<html xmlns="http://www.w3.org/1999/xhtml" xmlns:epub="http://www.idpf.org/2007/ops" xml:lang="en">
+  <head>
+    <title>Processing the 'direction' attribute in SVG content</title>
+  </head>
+  <body>
+    <p>Test passes if the next page (in SVG) display the text as follows:</p>
```
dlazin, line 6:
```suggestion
    <p>Test passes if the next page (an SVG) displays the text as follows:</p>
```

tests/cnt-i18n_svg-direction/EPUB/content_001.xhtml
```diff
@@ -0,0 +1,9 @@
...
+    <img src="imgs/valid_farsi_text.png" alt="Image version of the farsi text appearing in the SVG content document that follows this, and figuring the proper rendering (in farsi) of 'داستان SVG 1.1 SE طولا ني است.'"/>
```
dlazin, line 7:
```suggestion
    <img src="imgs/valid_farsi_text.png" alt="Image version of the Farsi text appearing in the SVG content document that follows this, and figuring the proper rendering (in Farsi) of 'داستان SVG 1.1 SE طولا ني است.'"/>
```

tests/cnt-i18n_svg-direction/EPUB/content_001.xhtml
```diff
@@ -0,0 +1,9 @@
+<html xmlns="http://www.w3.org/1999/xhtml" xmlns:epub="http://www.idpf.org/2007/ops" xml:lang="en">
```
dlazin, line 1: For consistency with pkg-dir_* and #confreq-rs-svg-direction, I think this should be named cnt-svg-direction.

Remember, underscores are only for if you have multiple tests for one normative statement. (I realize that no one understands this and it's arbitrary and I would totally change it if we could, but too late!)

tests/cnt-i18n_xhtml-dir/EPUB/content_001.xhtml
```diff
@@ -0,0 +1,22 @@
+<html xmlns="http://www.w3.org/1999/xhtml" xmlns:epub="http://www.idpf.org/2007/ops" xml:lang="en" dir="rtl">
```
dlazin, line 1: Can you name this cnt-xhtml-dir, please?

I see that I named the anchor #confreq-rs-html-dir, and it was a weird choice to omit the "x"; feel free to update the spec to make it use "xhtml" too. It's a new anchor so no one should be using it yet.

tests/cnt-i18n_xhtml-dir/EPUB/content_001.xhtml
```diff
@@ -0,0 +1,22 @@
...
+  <head>
+    <title>Processing the 'dir' attribute in XHTML content</title>
+  </head>
+  <body>
+    <div dir="ltr">
+      <p>Test passes if:</p>
+      <ul>
+        <li>the numbered list below is adjusted to the right edge of the screen</li>
```
dlazin, line 9: "adjusted" is a weird verb here. Do you mean "the numbered list is right-aligned" (or "aligned right")?

tests/cnt-i18n_xhtml-dir/EPUB/content_001.xhtml
```diff
@@ -0,0 +1,22 @@
...
+        <li>the numbering of the items are on the right of the corresponding texts</li>
```
dlazin, line 10:
```suggestion
        <li>the numerals appear to the right of the corresponding text</li>
```

tests/cnt-i18n_xhtml-dir/EPUB/content_001.xhtml
```diff
@@ -0,0 +1,22 @@
...
+        <li>the item with the ABC is rendered "backwards", i.e., as "zyxwv…"</li>
```
dlazin, line 11:
```suggestion
        <li>the last item is rendered as the alphabet backwards, i.e. as "zyxwv…"</li>
```

tests/cnt-i18n_xhtml-dir/EPUB/content_001.xhtml
```diff
@@ -0,0 +1,22 @@
...
+      </ul>
+    </div>
+
+    <ol>
+      <li>lorem ipsum dolor sit amet</li>
+      <li>consectetur adipiscing elit</li>
+      <li><bdo dir="rtl">abcdefghijklmnopqrstuvwxyz</bdo></li>
```
dlazin, line 18: I think it's fine to test this, but my reading of the spec is that it applies only to document-level dir, not element-level: "A reading system MUST process, for all [publication resources], the attributes that set the base direction for the documents' contents."

I assume that "base" means document-level; IMO <bdo dir="rtl"> is just direction, not "base direction" :)

Your discretion as to whether to keep this.

tests/cnt-i18n_xhtml-dir/EPUB/package.opf
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<package xmlns="http://www.idpf.org/2007/opf" xmlns:epub="http://www.idpf.org/2007/ops" version="3.0" xml:lang="en" unique-identifier="pub-id">
+  <metadata xmlns:dc="http://purl.org/dc/elements/1.1/">
+    <dc:coverage>Internationalization</dc:coverage>
+    <dc:creator>Ivan Herman</dc:creator>
+    <dc:date>2022-05-25</dc:date>
+    <dc:description>The content document's text direction is set to 'rtl' using the 'dir' attribute; the rendering should reflect this setting.</dc:description>
```
dlazin, line 7:
```suggestion
    <dc:description>The XHTML content document's text direction is set to 'rtl' using the 'dir' attribute; the rendering should reflect this setting.</dc:description>
```

tests/cnt-i18n_xhtml-lang/EPUB/content_001.xhtml
```diff
@@ -0,0 +1,12 @@
+<html xmlns="http://www.w3.org/1999/xhtml" xmlns:epub="http://www.idpf.org/2007/ops" lang="fr">
+  <head>
+    <title lang="en">Processing the 'lang' attribute in XHTML content</title>
+    <meta charset="utf-8"/>
+  </head>
+  <body>
+    <p lang="en">Test passes if the following quotation is surrounded by the pair of '«' and '»' characters (used in French), instead of the '“' and '”' quotation marks (used in English).</p>
```
dlazin, line 7:
```suggestion
    <p lang="en">Test passes if the following quotation is surrounded by a pair of '«' and '»' characters (used in French), instead of the '“' and '”' quotation marks (used in English).</p>
```

tests/cnt-i18n_xhtml-lang/EPUB/content_001.xhtml
```diff
@@ -0,0 +1,12 @@
+<html xmlns="http://www.w3.org/1999/xhtml" xmlns:epub="http://www.idpf.org/2007/ops" lang="fr">
```
dlazin, line 1: I would call this cnt-xhtml-lang, please :)

tests/cnt-i18n_xhtml-xml-lang/EPUB/content_001.xhtml
```diff
@@ -0,0 +1,12 @@
+<html xmlns="http://www.w3.org/1999/xhtml" xmlns:epub="http://www.idpf.org/2007/ops" xml:lang="fr">
```
dlazin, line 1: I would call this cnt-xml-lang or cnt-xhtml-xml-lang (but IMO that's redundant; XHTML is XML).

tests/cnt-i18n_xhtml-xml-lang/EPUB/content_001.xhtml
```diff
@@ -0,0 +1,12 @@
...
+  <head>
+    <title lang="en">Processing the 'xml:lang' attribute in XHTML content</title>
```
dlazin, line 3:
```suggestion
    <title lang="en">Processing the 'xml:lang' attribute in XML content</title>
```

tests/cnt-i18n_xhtml-xml-lang/EPUB/nav.xhtml
```diff
@@ -0,0 +1,12 @@
+<html xmlns="http://www.w3.org/1999/xhtml" xmlns:epub="http://www.idpf.org/2007/ops" xml:lang="en">
+  <head>
+    <title>Processing the 'xml:lang' attribute in XHTML content</title>
```
dlazin, line 3:
```suggestion
    <title>Processing the 'xml:lang' attribute in XML content</title>
```

tests/cnt-i18n_xhtml-xml-lang/EPUB/package.opf
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<package xmlns="http://www.idpf.org/2007/opf" xmlns:epub="http://www.idpf.org/2007/ops" version="3.0" xml:lang="en" unique-identifier="pub-id">
+  <metadata xmlns:dc="http://purl.org/dc/elements/1.1/">
+    <dc:coverage>Internationalization</dc:coverage>
+    <dc:creator>Ivan Herman</dc:creator>
+    <dc:date>2022-05-24</dc:date>
+    <dc:description xml:lang="en">
+      The content document's language is set to 'fr' using the 'xml:lang' attribute; the rendering should reflect this setting.
```
dlazin, line 8:
```suggestion
      The XHTML content document's language is set to 'fr' using the 'xml:lang' attribute; the rendering should reflect this setting.
```

tests/cnt-i18n_xhtml-xml-lang/EPUB/package.opf
```diff
@@ -0,0 +1,26 @@
...
+    </dc:description>
+    <dc:identifier id="pub-id">cnt-i18n_xhtml-xml-lang</dc:identifier>
+    <dc:language>en</dc:language>
+    <dc:publisher>W3C</dc:publisher>
+    <dc:title>Processing the 'xml:lang' attribute in XHTML content</dc:title>
```
dlazin, line 13:
```suggestion
    <dc:title>Processing the 'xml:lang' attribute in XML content</dc:title>
```

tests/cnt-i18n_xhtml-xml-lang/EPUB/package.opf
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
dlazin, line 1: This test only addresses XHTML content, but the spec says that xml:lang must be tested in the package etc too. Should that also be in this test? (nav.xhtml too?) Or is that already covered in one of your existing tests, which just needs to be tagged to https://w3c.github.io/epub-specs/epub33/rs/#confreq-rs-xml-lang?

tests/cnt-i18n_xhtml-lang/EPUB/package.opf
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<package xmlns="http://www.idpf.org/2007/opf" xmlns:epub="http://www.idpf.org/2007/ops" version="3.0" xml:lang="en" unique-identifier="pub-id">
+  <metadata xmlns:dc="http://purl.org/dc/elements/1.1/">
+    <dc:coverage>Internationalization</dc:coverage>
+    <dc:creator>Ivan Herman</dc:creator>
+    <dc:date>2022-05-24</dc:date>
+    <dc:description xml:lang="en">
+      The content document's language is set to 'fr' using the 'lang' attribute; the rendering should reflect this setting.
```
dlazin, line 8:
```suggestion
      The XHTML content document's language is set to 'fr' using the 'lang' attribute; the rendering should reflect this setting.
```

dlazin (approved):
Great tests! Thank you.

iherman:
Closing this PR without merging by virtue of https://github.com/w3c/epub-specs/pull/2330

PR #152, opened 2022-05-27
Remove SHOULD marker from tests that are MUSTs under media overlays

dlazin:
As near as I can tell, these are all the MO tests that correspond to a MUST attribute (but see my question in Slack). As discussed in https://github.com/w3c/epub-tests/pull/148 and the last meeting, we have decided that these should be marked as MUST, because CR proof is more important than telegraphing to an RS whether they ought to pass a test.


iherman:
For the records, here is an answer to your question on slack: https://w3ccommunity.slack.com/archives/C02KKMQH50S/p1653722260554639

TL;DR both `mol-embed` and `mol-embed_fxl` should be _MUST_ tests in my view.

iherman (approved):

PR #157, opened 2022-05-31
Automatically generate and OPDS file for the tests

iherman:
[OPDS](https://drafts.opds.io/opds-2.0.html) is a draft specification for epub catalogues. Its primary usage, for the testing context, is to provide a catalogue file that all Readium based reading systems (and maybe others) can use to automatically load all the tests into the system. Generating this catalogue by the automatic tooling is easy, because all the required data are available, it is just a matter of generating yet another file in the repository.

Reviewers, instead of looking at the typescript code, the best is to look at the generated [OPDS Index for the tests](https://raw.githubusercontent.com/w3c/epub-tests/generate_opds/docs/opds/opds.json) to see if it is correct (per OPDS).


src/main.ts
```diff
@@ -2,10 +2,11 @@ import { argv } from "process";
 import * as fs_old_school from "fs";
 const fs = fs_old_school.promises;
 
-import { ReportData, ImplementationReport, Constants } from './lib/types';
-import { get_report_data, get_template } from "./lib/data";
+import { TestData, ReportData, ImplementationReport, Constants } from './lib/types';
+import { get_test_metadata, get_report_data, get_template } from "./lib/data";
```
dlazin, line 6: Super-nit: since lines 5 and 27 refer to test_data and TestData, and I assume all 3 references are related, can we call this function get_test_data?

dlazin (approved):
Cool! I bet I will want to consume this JSON file in some other way later :)

src/lib/opds.ts
```diff
@@ -0,0 +1,104 @@
...
+        }]
+        const metadata: PublicationMetadata = {
+            "@type"     : 'http://schema.org/Book',
+            identifier  : `${Constants.TEST_URL_BASE}/${test.identifier}`,
+            title       : test.title,
+            author      : test.creators.length === 1 ? test.creators[0] : test.creators,
+            description : test.description,
+            collection  : test.coverage,
+            modified    : test.modified,
+            publisher   : 'W3C',
```
danielweck, line 80: Bonjour Ivan :)

Could you please add:
```
subject: ['epub-test'],
```

...this way, importing publications from this OPDS feed will automatically inherit the `epub-test` "tag" which will be super useful in Thorium (for example), to triage the test suite using search/filter/sort in our v2.0 bookshelf (when running the tests, we will manually add our own tags, e.g. `epub-test-fail` / `epub-test-pass`, `epub-test-todo`, etc.).

Note that the HTTP content-type header for the OPDS feed is incorrect:
* https://raw.githubusercontent.com/w3c/epub-tests/generate_opds/docs/opds/opds.json
...so instead we use this URL in Thorium:
* https://raw.githack.com/w3c/epub-tests/generate_opds/docs/opds/opds.json

Also note that internally we occasionally use a online "micro service" which conveniently visualises a Readium Web Pub Manifest or ODPS feed, and more crucially that runs JSON Schema validation (scroll to the very bottom to see the result):
* https://readium2.herokuapp.com/opds-v2-browse/https%3A%2F%2Fraw.githack.com%2Fw3c%2Fepub-tests%2Fgenerate_opds%2Fdocs%2Fopds%2Fopds.json

RWPM `subject`:
* https://github.com/readium/webpub-manifest/tree/master/contexts/default#subjects
JSON Schema:
* https://github.com/readium/webpub-manifest/blob/ee2491413fe33133c42569c61e9aa1a51deb0d52/schema/metadata.schema.json#L101-L103
* https://github.com/readium/webpub-manifest/blob/ee2491413fe33133c42569c61e9aa1a51deb0d52/schema/subject.schema.json

Concrete real-world use of `subject` at Feedbooks (note the `links` property which allows "clicking" on a subject tag in the user interface to fetch an OPDS feed URL that shows only publications with the selected subject):

* http://readium2.herokuapp.com/opds-v2-browse/https%3A%2F%2Fcatalog.feedbooks.com%2Fitem%2F2159345.json
* https://catalog.feedbooks.com/item/2159345.json

src/lib/opds.ts
```diff
@@ -0,0 +1,104 @@
...
+
+interface PublicationMetadata {
+    "@type": string;
+    identifier: string;
+    title: string;
+    author: string | string[];
+    description: string;
+    collection: string;
+    modified: string;
+    publisher: string
```
danielweck, line 23: Hello Ivan, followin-up on my first long comment:
please add `subject: string | string[];`

src/lib/opds.ts
```diff
@@ -0,0 +1,104 @@
...
+        title: string;
+    };
+    links : PublicationLink[];
+    publications: Publication[];
+}
+
+export function create_opds(tests: TestData[]): OPDS {
+
+    /** All cover images are identical */
+    const images: ImageLink[] = [{
```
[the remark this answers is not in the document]
iherman: These will be alive once the PR is merged; these files are not yet on the main branch.

src/lib/opds.ts
iherman: > Note that the HTTP content-type header for the OPDS feed is incorrect:
> 
> * [`generate_opds`/docs/opds/opds.json (raw)](https://raw.githubusercontent.com/w3c/epub-tests/generate_opds/docs/opds/opds.json)
>   ...so instead we use this URL in Thorium:
> * [raw.githack.com/w3c/epub-tests/generate_opds/docs/opds/opds.json](https://raw.githack.com/w3c/epub-tests/generate_opds/docs/opds/opds.json)

Just to check... Isn't this also a problem with the fact that this file is not yet on the main branch; it will be there when the PR is merged. Or is there an entry in the HTTP header that you expect and that the raw github response does not provide?

danielweck: IIRC the HTTP content-type for JSON will be emitted correctly if/when serving files from gh-pages or another publishing branch in the GitHub repository. However with all other branches it is served as plain text, (or octet stream, i cannot remember and I am on my phone soncannot eailsiky check ;) ... thus the use of GitHack or other suitablr CDN.

iherman: Note the previous comment: there is another json file that is on the main branch, with the URI https://w3c.github.io/epub-tests/drafts/config.json. The HTTP headers are:

```
HTTP/2 200 
server: GitHub.com
content-type: application/json; charset=utf-8
permissions-policy: interest-cohort=()
x-origin-cache: HIT
last-modified: Sat, 28 May 2022 17:12:37 GMT
access-control-allow-origin: *
etag: "62925805-291"
expires: Wed, 08 Jun 2022 06:30:48 GMT
cache-control: max-age=600
x-proxy-cache: MISS
x-github-request-id: 7FA0:9420:988E6F:A3D4D9:62A03FC0
accept-ranges: bytes
date: Wed, 08 Jun 2022 06:20:48 GMT
via: 1.1 varnish
age: 0
x-served-by: cache-fco2270023-FCO
x-cache: MISS
x-cache-hits: 0
x-timer: S1654669248.401139,VS0,VE104
vary: Accept-Encoding
x-fastly-request-id: f23f77ec52c3771daed5e9a10407efa706325b35
content-length: 657
```

I would think that should work with thorium as well...

danielweck: that's because it is a GitHub.io URL (i.e. the source is a GitHub publishing branch, regardless of its symbolic name, equivalent to gh-pages).

src/lib/opds.ts
iherman: done.

src/lib/opds.ts
iherman: > that's because it is a GitHub.io URL (i.e. the source is a GitHub publishing branch, regardless of its symbolic name, equivalent to gh-pages).

Right. But once this PR is published, https://w3c.github.io/epub-tests/opds/opds.json will behave exactly the same way.

Ie, the overall link for the opds.json file should remain the same imho...

iherman: B.t.w., I have done the changes for the subject, but still have to regenerate the files.

iherman: > still have to regenerate the files

Also done...

danielweck: Thank you Ivan, it works great. See screenshots (ignoring the broken cover images, as previously discussed):
Screenshot 2022-06-08 at 07 51 00
Screenshot 2022-06-08 at 07 50 18
Screenshot 2022-06-08 at 07 50 02

danielweck: Now that the `epub-test` is automatically set in the local bookshelf when importing EPUBs from the OPDS feed, I can easily search/filter/sort in the "data grid" view, to show only the EPUB test suite, and/or to display only my `epub-test-pass` or `epub-test-fail` tags (or even `epub-test-question` etc. if I am unsure about a test):

Screenshot 2022-06-08 at 07 56 19

iherman: Wow. This all looks great!

I will add some words about the availability of this file in the general description. Can we expect additional review? @llemeurfr referred to Pierre possibly reviewing this... If so, I am happy to wait with merging.

danielweck: Well, the OPDS feed "works" on multiple levels:
* We can open it in Thorium, and download publications with automatic "tag" import (`subject` metadata). The broken cover image is not critical, and it will be fixed when merged anyway (if I understand correctly).
* This OPDS feed is in the v2 JSON format, which validates against the official JSON Schema (scroll all the way to the bottom of the page to see the result): https://readium2.herokuapp.com/opds-v2-browse/https%3A%2F%2Fraw.githack.com%2Fw3c%2Fepub-tests%2Fgenerate_opds%2Fdocs%2Fopds%2Fopds.json
* Alternative online validator link: https://www.jsonschemavalidator.net/s/GIauaQAy (JSON Schema source: https://raw.githubusercontent.com/opds-community/drafts/master/schema/feed.schema.json )

...so, as far I we are concerned, this PR can be merged. We will report back if there are issues.

llemeurfr: Daniel has carefully checked the file, therefore Pierre's review is not required anymore. Great work! that will be also useful for checking the test suite in many mobiles apps (ex. Aldiko).

iherman: Ok! Merging now.

danielweck:
Thank you Ivan. I tested the URL https://w3c.github.io/epub-tests/opds/opds.json
...the cover images now work in the OPDS feed (see screenshots).

Screenshot 2022-06-08 at 16 26 22

Screenshot 2022-06-08 at 16 26 44

danielweck:
Note that once imported from the OPDS feed into the local bookshelf, the test EPUBs don't actually have a cover image (see screenshots below). You can leave it this way and enjoy the fact that in many reading systems, publication titles are usually displayed in place of the missing cover image ... or you can add the same pretty cover image for all test EPUBs, which will then all have identical visuals when presented in the library bookshelf. I have no preference :)

Screenshot 2022-06-08 at 16 27 35

Screenshot 2022-06-08 at 16 37 02

Screenshot 2022-06-08 at 16 36 21

iherman:
Yes, we have never added cover images to the tests. Maybe once, when I will have a lot of time, I will create a script that adds those to the tests, modifying some SVG file that I use now in the OPDS metadata. But it is not that important and does not have a high priority...

iherman:
Just out of curiosity, @danielweck: does Readium/Thorium has an update mechanism for OPDS? The way it works now is that if new tests are added to the test suite via a PR, the corresponding github action would regenerate the OPDS file with the new tests added. How would you guys use that in your workflow?

I do not know whether there is a notification mechanism in github that could ping you if such a change occurred (except 'watching' the repo for new PR merges).

gautierchomel:
As OPDS displays _Title_ of the File but [Report](https://github.com/w3c/epub-tests/blob/main/reports/xx-template.json) refers to file _Name_ there is some complexity to make correspondences. 

Numbering the tests would help.

iherman:
Numbering is a bit complicated because the tests are changing all the time (new are added), some are developed in parallel, etc.

In fact, the report file uses the _unique identifier_ of the test (and corresponds to the identifier entry in the package file's metadata).

The OPDS specification requires the `identifier:` field to be a URL, which is the URL for the unzipped version of the test, which is the concatenation of a fix URL and the test identifier. Is the requirement for the OPDS `identifier` value being a URL a strict one? Because the generation script could certainly simply copy the test identifier.

In any case, the correspondence between the OPDS and the report is the test identifier value.

danielweck:
Thorium currently doesn't display dc:identifier metadata (neither in the local bookshelf nor in the OPDS explorer). This is a design decision, not a bug ... but we could of course revisit this.

iherman:
Well... I could of course add the identifier value to the title (e.g., add it in parentheses at the end) in the OPDS file if that helps. (Although I am not sure it does, it fact.) 

Having a way in Thorium to make the identifier available to the user would certainly be the simplest.

PR #167, opened 2022-06-17
Removing the acid test

### LINKED ISSUE #164: The "acid" test is not specific enough (not "unit" testing) + fails with typical stylesheet overrides from reading system "user agent" and/or user formatting preferences
Screenshot from Apple iBooks / MacOS Books.app (same broken rendition with Thorium):

Screenshot 2022-06-15 at 14 34 52

* Direct HTML link (successful rendering in a "vanilla" web browser, with default user agent stylesheet and no major user styling customisations): https://raw.githack.com/w3c/epub-tests/main/tests/cnt-css-support/EPUB/content_001.xhtml#top
* EPUB download link: https://github.com/w3c/epub-tests/blob/main/tests/cnt-css-support.epub

The rationale for including the Acid test in the EPUB test suite was mentioned in this PR https://github.com/w3c/epub-tests/pull/7

> I also made a version of the Acid2 test, as that seemed one way to test the requirement about reading systems with a viewport: "If it has a Viewport, it MUST support visual rendering of XHTML Content Documents as defined in § 4.3.1 CSS Conformance."

This Acid test was indeed a useful exercise that offered a vivid visual demonstration of the shortcomings of some competing CSS implementations in web browsers and other rendering engines. However I think the value of Acid test in the context of the EPUB test suite is arguably debatable because it is not clear what feature is tested exactly. Furthermore, reading systems typically apply a default "user agent" stylesheet (+ additional overrides based on user preferences) that cause different rendering than vanilla web browsers. For example, pagination using CSS columns usually involves overriding margins, etc.

My recommendation would be to eliminate the Acid test from this test suite.

iherman:
Fix #164


iherman:
There is a resolution on removing this; merging.

PR #173, opened 2022-06-23
Adding a hreflang test

iherman:
The PR adds a tests to the link element in [§5.3](https://w3c.github.io/epub-specs/epub33/rs/#sec-pkg-doc-metadata) of the reading system spec, for the following statement:

> The language identified in an hreflang attribute is purely advisory. Upon fetching the resource, a reading system MUST use the language information associated with the resource to determine its language, not the metadata included in the link to the resource.

This requires careful review. I rely on the possibility of using a `<link>` to define a description:

```xml
<link rel="dcterms:description" href="description.xhtml" 
   media-type="application/xhtml+xml" hreflang="fr"/>
```

and checking whether the display of the result is done in the 'language unknown' or the French way. I hope that is correct, although two things bother me:

- (obviously) I do not know whether we have any RS that would use the linked description instead of the one given in the package file
- even if the intention is there for the RS, the package description is specified via `dc:description`, but `dc` is not allowed for the value of `rel` in the `link` element, only `dcterms:description`. Technically, these two are different.

I am all open to use a different example for `hreflang` testing, but I ran out of ideas...

(The anchors in the spec are not yet set.)

### DIFF
--- a/tests/pkg-linked-records_hreflang/EPUB/content_001.xhtml
+++ b/tests/pkg-linked-records_hreflang/EPUB/content_001.xhtml
@@ -1,0 +1,15 @@
+<html xmlns="http://www.w3.org/1999/xhtml" xmlns:epub="http://www.idpf.org/2007/ops" xml:lang="en">
+  <head>
+    <title>Link to a description with hreflang</title>
+  </head>
+  <body>
+    <h1>The <code>hreflang</code> attribute must be ignored when processing the target of a <code>link</code> element</h1>
+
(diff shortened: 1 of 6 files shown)


mattgarrish:
> I do not know whether we have any RS that would use the linked description

I haven't heard of a reading system that uses linked records, let alone individual properties, let alone individual properties that would take parsing their value out of a grammar to obtain, so I suspect there will never be implementations to test.

I wonder if this is a case for a lowercase "must", since it feels oddly familiar to what we just discussed with respect to SVG. We're saying that the language specification in the package document is advisory and that's all that matters from an EPUB perspective. Do we need to require that reading systems obey the language specification rules of the host format, or is that implicit in the processing of any linked data?

mattgarrish:
This is more what I'm thinking it should be:

> The language identified in an hreflang attribute is purely advisory. The language information expressed in the resource determines its language, not the metadata included in the link to the resource.

PR #182, opened 2022-07-26
Added two more tests to complete the ICB setting rules

iherman:
Title tells it all...

Anchors are added via PR https://github.com/w3c/epub-specs/pull/2373


iherman:
@wareid we know that Dan is less available these days; could you review these tests to see if I made some obvious mistake?

PR #184, opened 2022-07-27
Spine rendition override tests

### LINKED ISSUE #183: Taking care of spine override tests
Ie, https://w3c.github.io/epub-specs/epub33/rs/#spine-overrides

iherman:
Fix #183

@wareid @dauwhe @mattgarrish I would really need a review of this. These two tests sound correct to me, but they are not implemented neither by Thorium nor Apple Books, so we have to be sure it reflects the intention of the spec.


iherman:
Anchors and references are in https://github.com/w3c/epub-specs/pull/2374

iherman:
@mattgarrish could you have a quick look at these tests? I would like to be sure I did not misunderstand the spec as none of the RS-s on my machine pass them...

mattgarrish:
The tests look fine to me. Not completely surprising that the behaviours aren't implemented. That's always been the knock on FXL. We introduced a lot of properties and values and RSes generally only implemented the most interesting to them.

There's probably not a lot of people overriding FXL publications to add reflowable documents. The more common case is to override a reflowable to add an FXL cover page, plates, or things of that nature. If you just want to get a positive result, I suspect if you turn the overriding around you'll get at least some passes.

That reading systems aren't checking for conflicting overrides isn't too surprising, either, so I doubt there's any change to the test that would get a different result.

iherman:
> The tests look fine to me. Not completely surprising that the behaviours aren't implemented. That's always been the knock on FXL. We introduced a lot of properties and values and RSes generally only implemented the most interesting to them.
> 
> There's probably not a lot of people overriding FXL publications to add reflowable documents. The more common case is to override a reflowable to add an FXL cover page, plates, or things of that nature. If you just want to get a positive result, I suspect if you turn the overriding around you'll get at least some passes.

Ah.

I will try to add a test next week along those lines by switching the override around. But I would keep these.

Next week, probably not before Tuesday.

iherman:
@mattgarrish I have added a new test with reflowable as global setting and pre-paginated as an override value. This test indeed works on Apple Books (but it does not on Thorium :-(.

A quick check before I merge this? Thanks.

mattgarrish:
Ya, the new test looks fine to me.

PR #195, opened 2022-08-03
Scripting-content-fallback

iherman:
Test for the last statement in [§6.4](https://w3c.github.io/epub-specs/epub33/rs/#sec-scripted-content) on following a fallback for scripted content.

The main difficulty of the test is to find a RS that does not support scripting...

Spec counterpart PR is in https://github.com/w3c/epub-specs/pull/2389


mattgarrish:
Google Play Books doesn't support scripting, so you could load it there and see what happens. Otherwise, this just looks like a straightforward manifest fallback test based on the scripted property presence, so nothing to add from me.

mattgarrish (approved):

mattgarrish:
We can skip testing html fallbacks, like for canvas, as they're defined by html, correct?

iherman:
> We can skip testing html fallbacks, like for canvas, as they're defined by html, correct?

I would think so.

iherman:
> Google Play Books doesn't support scripting, so you could load it there and see what happens. Otherwise, this just looks like a straightforward manifest fallback test based on the scripted property presence, so nothing to add from me.

I do not have Google Play Books installed, so I cannot check. But, at this point, the really important question is whether the test itself is correct or not. If it is, the question will be whether this feature (and fallback in general) would become under-implemented...

mattgarrish:
> I do not have Google Play Books installed

It's web based, so if you have a google account you should be able to load the book through their online interface: https://play.google.com/books

There's an option to upload files in the top menu bar.

mattgarrish:
> the question will be whether this feature (and fallback in general) would become under-implemented

Well, ya, that's what my comment was getting at. It's a specific kind of manifest fallback, but if manifest fallbacks aren't supported then this won't be by default. They're pretty tightly bound, and I'd expect `properties` fallbacks will have less support then `media-type` ones.

Issue #197, opened 2022-08-09
Completing  Loading the media overlay tests


iherman:
I take care of completing the missing tests in [§9 Loading the media overlay](https://w3c.github.io/epub-specs/epub33/rs/#sec-behaviors-loadings)

PR #198, opened 2022-08-09
Tests for the "Loading the media overlay" section

iherman:
Fix: #197

Adding a MUST and a SHOULD test.

- The MUST test starts the audio playback in the middle of the audio track, corresponding to the statement "Note that the start of an EPUB content document could correspond to an element at the start or in the middle of a media overlay."
- The SHOULD test's audio playback is used in two consecutive content files; the expectation is that the RS switches from the first to the second automatically as the read loud goes on.

Also, the test in the second sentence (`mol-support_xhtml`) can also be seen as a test for the very first sentence in the paragraph; the spec references have been modified accordingly.

Corresponding spec PR is https://github.com/w3c/epub-specs/pull/2395


tests/mol-support_xhtml-load-next/EPUB/content_001.xhtml
```diff
@@ -0,0 +1,19 @@
+<html xmlns="http://www.w3.org/1999/xhtml">
+<head>
+    <meta charset="utf-8" />
+    <title>MO playback switching XHTML content</title>
+</head>
+<body>
+    <p>Test passes (i.e., the Reading System supports playback with XHTML Media document) if</p>
+    <ol>
+        <li>The next play is properly displayed as an XHTML content</li>
```
marisademeglio, line 9: "The next page"?

marisademeglio:
Can the MUST test start in the middle of the MO document? So that the reading system has to search the MO structure for the text element ID and start playback midway through? We already test seeking to an audio file offset (I think it's one of the first MO tests) so I don't think that's useful here. 

>>Also, the test in the second sentence (mol-support_xhtml) can also be seen as a test for the very first sentence in the paragraph; the spec references have been modified accordingly.

So the test EPUB `mol-support_xhtml` tests two things? What sentences and what paragraphs are you referring to? In the spec, in the test book, elsewhere?

tests/mol-support_xhtml-load-next/EPUB/content_001.xhtml
iherman: Oops :-) Thanks!

iherman:
> Can the MUST test start in the middle of the MO document? So that the reading system has to search the MO structure for the text element ID and start playback midway through? We already test seeking to an audio file offset (I think it's one of the first MO tests) so I don't think that's useful here.

Technically, you are right: most of the other tests show that seeking into the audio file is done so, from that point of view, this test does not add too much. However, the formal requirement we have, from a CR point of view, is that every _MUST_ statement must have a corresponding test. I believe most of the tests are constructed in a way that they _start_ the read aloud at the beginning of the audio content, and they usually do other things as well. The only feature of this test is to _start_ with an offset, which is what the spec statement says. In a way, this is the reflection of the fact that this test suite has a very fine granularity.

Again, from an implementation point of view it does not add anything; from the spec review point of view its presence mean we do not have to add additional explanations...

> 
> > > Also, the test in the second sentence (mol-support_xhtml) can also be seen as a test for the very first sentence in the paragraph; the spec references have been modified accordingly.
> 
> So the test EPUB `mol-support_xhtml` tests two things? What sentences and what paragraphs are you referring to? In the spec, in the test book, elsewhere?

Sorry I was not clear. I refer to the first sentence in §9.1:

> When a reading system loads a [package document](https://www.w3.org/TR/epub-33/#dfn-package-document), it MUST refer to the [manifest](https://www.w3.org/TR/epub-33/#dfn-epub-manifest) [item](https://www.w3.org/TR/epub-33/#dfn-item) elements' [[epub-33](https://w3c.github.io/epub-specs/epub33/rs/#bib-epub-33)] media-overlay attributes to discover the corresponding media overlays for [EPUB content documents](https://www.w3.org/TR/epub-33/#dfn-epub-content-document).

_Any_ MO test covers this requirement but we are a bit in the same situation as above: we should assign a test that may also test this statement. So the simplest is to reuse the same test used in this (short) paragraph.

marisademeglio:
> The only feature of this test is to start with an offset, which is what the spec statement says. In a way, this is the reflection of the fact that this test suite has a very fine granularity.

If this is the spec statement:

> Playback MUST start at the media overlay element which corresponds to the desired EPUB content document starting point. Note that the start of an EPUB content document could correspond to an element at the start or in the middle of a media overlay.

To me, it means to start in the middle of the media overlay document (the SMIL file, not necessarily the audio file). 

And, I remember that this statement is in the spec because we know it's easy for implementations to overlook this and just default to starting at the beginning of the SMIL file without fast forwarding through the tree structure to a specific element reference.

iherman:
> 
> If this is the spec statement:
> 
> > Playback MUST start at the media overlay element which corresponds to the desired EPUB content document starting point. Note that the start of an EPUB content document could correspond to an element at the start or in the middle of a media overlay.
> 
> it means to start in the middle of the media overlay document (the SMIL file, not necessarily the audio file).
> 
> And, I remember that this statement is in the spec because we know it's easy for implementations to overlook this and just default to starting at the beginning of the SMIL file without fast forwarding through the tree structure to a specific element reference.

Argggh! I completely misunderstood. My bad.

Just to make it clear then: the SHOULD test (ie, mol-support_xhtml-load-next) is actually testing this, right? It has two content documents sharing a single MO file; the second content document's audio counterpart is governed by a separate <seq> in the "middle" of the MO file. 

I used it as a test for the SHOULD statement, whereby there is a smooth continuation of the text read out, but it seems that this is, in fact, the real test to be used. The current MUST test is indeed to be thrown out...

I would have to reformulate the tests accordingly then...

iherman:
@marisademeglio I have removed the incriminated test. The new test has a different structure and, I believe, can be considered as a test for the statement we were discussing.

PR #199, opened 2022-08-09
Changed the test descriptions

### LINKED ISSUE #151: basic MO audio rendering test tests more than basic MO audio rendering
In addition to if an audio clip is played from clip start to clip end, it also tests if the text is highlighted and if the reading system has a button to start playback (which is technically not a requirement, if the RS just started playback automatically we would still say it supports MO).

The relevant commit is https://github.com/w3c/epub-tests/commit/631baf09fe6eb2003703e29092e8063e9cbf2f06

iherman:
This is to fix #151 : I have changed the test descriptions by putting a separate statement of the form:

> The Reading System may highlight the text being read with a green background with the text grayed out; but this feature is not required for the test to pass.

and removing (if applicable) the reference to the highlights from the bullet items describing the success criteria of the tests.


iherman:
I will merge this. It is only aesthetic anyway, it does not change the tests themselves. On the other hand, it changes a lot of files, and it can lead to horrific merge conflicts with real test developments...

PR #201, opened 2022-08-30
Added a test for font obfuscation

### LINKED ISSUE #39: Test obfuscated resources (fonts)
I will test the following normative statement:

https://www.w3.org/TR/epub-rs-33/#confreq-ocf-fobfus
Reading Systems MUST support deobfuscation of resources as defined in Resource Obfuscation [EPUB-33].

iherman:
@dlazin I created a test to supersede the draft PR #21, which has been pending as a draft for very long due to the uncertainty of obfuscation. As an information, I used Sigill to obfuscate a font manually; the resulting font is certainly obfuscated but also recognized by Thorium and Apple Books (ie, both passes this test), so we can be fairly sure the obfuscation is indeed as it should be.

Fix #39


dlazin (approved):
Thank you, Ivan! I really appreciate this. I will revert my pending PR, of course.

tests/ocf-font_obfuscation/EPUB/package.opf
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<package xmlns="http://www.idpf.org/2007/opf" xmlns:epub="http://www.idpf.org/2007/ops" version="3.0" xml:lang="en" unique-identifier="pub-id">
+  <metadata xmlns:dc="http://purl.org/dc/elements/1.1/">
+    <dc:coverage>Open Container Format</dc:coverage>
+    <dc:creator>Ivan Herman</dc:creator>
+    <dc:date>2022-08-30</dc:date>
+    <dc:description>An obfuscated (TrueType) font should be displayed after de-obfuscation.</dc:description>
+    <dc:identifier id="pub-id">ocf-font_obfuscation</dc:identifier>
```
dlazin, line 8: Oh, just to check, if you change the pub-id (only) and then regenerate the .epub, does the font stop working (in Apple Books, at least)? I think that's the confirmation I've used in the past.

iherman: Good point! I have just checked it on my machine and yes, the modified epub fails on Thorium. As it should.

I briefly considered adding that version to the test suite, but I hesitate. After all, such a "negative" test does not necessarily test a different aspect of the RS behaviour. On the other hand, it can be considered as a test for the correctness of the Sigil obfuscation algorithm, so we could take this `ocf-font_obfuscation_bis`, as I called it, as a test of the obfuscation algorithm proper in the core spec. (Remember, we are talking about a _SHOULD_ thing as far as the RS is concerned, so we are not under a pressure to provide some exhaustive algorithm or implementation report.)

Our choices are:

1. Leave it as is and merge
2. Add the "bis", and add a reference to both tests to the Core spec's [§4.3.3.](https://www.w3.org/TR/epub-33/#obfus-keygen)

dlazin: I like the idea of adding the second test. It's simple for us to do, and helpful for a RS to understand if they (somehow) don't respect the ID when deobfuscating.

PR #224, opened 2022-09-14
Updating reports template and contributing docs

wareid:
In response to some of the discussion from the WG F2F I've made some changes to the reporting template to include variant information, and to the contributing doc to add information on what variant should be used for.

### DIFF
--- a/reports/xx-template.json
+++ b/reports/xx-template.json
@@ -1,6 +1,7 @@
 {
     "name": "(Implementation's name)",
     "ref": "https://www.example.com",
+    "variant": "(Implementation OS, Version, and any other important details)",
     "tests": {
         "cnt-css-fonts": null,
         "cnt-mathml-support": null,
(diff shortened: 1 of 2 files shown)


reports/xx-template.json
```diff
@@ -1,6 +1,7 @@
 {
     "name": "(Implementation's name)",
     "ref": "https://www.example.com",
+    "variant": "(Implementation OS, Version, and any other important details)",
```
dlazin, line 4: This is good, but note that I attempted an empty string in https://github.com/w3c/epub-tests/commit/753181b19e7c72fa40b803ed6004da6bf0699230#diff-913f5d0dcb49595aa28e4880047c868ec239667b0280d4345467ec9335a1569b, and I don't see that on the left here. Is your branch out of date? I think you will have to merge, but I approve your version as being superior to mine.

dlazin (approved):

PR #226, opened 2022-09-15
Fix for the EPUB writing mode test

wareid:
Fixed issue where the test was difficult to run in a single page.


tests/css-epub-writing-mode/EPUB/content_002.xhtml
```diff
@@ -0,0 +1,11 @@
+<html xmlns="http://www.w3.org/1999/xhtml" xmlns:epub="http://www.idpf.org/2007/ops" xml:lang="en">
+  <head>
+    <title>EPUB CSS Writing Modes</title>
+  </head>
+  <body>
+    <h1>-epub-writing-mode Tests</h1>
+
+    <p style="margin: 1em; -epub-writing-mode: vertical-rl">2. これはテストです。This test passes if this text renders vertically starting from the right-hand side of the page.</p>
```
bduga, line 8: I think it would be better to put this on `head` or `body`, since the [principal writing mode](https://www.w3.org/TR/css-writing-modes-3/#principal-flow) for the document will be determined from there. Same for the next file.

PR #228, opened 2022-09-20
Renaming the layout tests

wareid:
Renaming the layout tests to match spec structure better.


iherman (commented):
I know it is a pain but… would it be possible to systematically change the `dcterms:modified` metadata entries?

Also, as a measure of courtesy to our testers, maybe we should also rename the existing test results' JSO files to get the rename done for them (and notifiy them about the change). The templates will be automatically updated by the script.

wareid:
The only reason I didn't change the modified field was because this change would not force a retest, since it was a simple renaming, but I can change it if needed. Can change the results files as well too.

iherman:
> The only reason I didn't change the modified field was because this change would not force a retest, since it was a simple renaming, 

That is correct. My only worry is whether we hopelessly mess up existing testers by making some tests disappear an some come out of the blue...

> Can change the results files as well too.

This would certainly help those that have already submitted some (we may want to warn them by email). I know at least one who has not submitted one so far (in edrlab) who should be notified, too...

iherman (approved):

iherman:
I have just approved it!

iherman:
@yanshin1499 I would like to understand what your comment was intended for. Did you forget to add a comment ot the quote?

PR #232, opened 2022-09-22
Update tests using SVG to be usable on bw displays

iherman:
In some cases a simple change of black or white bacground works. On other cases a solid background was replaced by some pattern.


wareid (approved):

iherman:
(Got the green light from @gregoriopellegrino on an email, merging)

PR #233, opened 2022-10-01
ID value for title

iherman:
After the [discussion on the F2F meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2022-09-13-epub#section2), followed by further discussions on email this PR changes the structure for _all_ tests by setting the `dc:title` element to the identifier value. This means that, when injected/sideloaded the tests are shown by the reading systems with their ID, which makes it easier for testers to report the test results.

The [index page](https://raw.githack.com/w3c/epub-tests/id-for-title/docs/drafts/index.html) of the tests have also changed accordingly; as a bonus, the space for the descriptions is slightly larger now.

Cc @bduga @davemanhall @gregoriopellegrino @dauwhe @dlazin @wareid


wareid (approved):

tests/pkg-unique-id_duplicate/EPUB/package.opf
```diff
@@ -11,17 +10,17 @@
     <dc:identifier id="pub-id">pkg-unique-id</dc:identifier>
     <dc:language>en</dc:language>
     <dc:publisher>W3C</dc:publisher>
-    <dc:title>Unique identifier reused</dc:title>
+    <dc:title>pkg-unique-id</dc:title>
```
davemanhall, line 13: pkg-unique-id_duplicate?

davemanhall (commented):
Fantastic - this will be a great help.  Looks like one title may need to change? pkg-unique-id_duplicate (Since the dc:identifier is the same between two documents?

tests/pkg-unique-id_duplicate/EPUB/package.opf
iherman: Ouch... I did not realize this test and I missed this... Thanks for noting this, @davemanhall!

PR #234, opened 2022-10-04
Updated the lay-fxl-icb* tests for bw displays

iherman:
These tests relied on two colored rectangles in xhtml (set via css); these tests may not have been usable for b/w reading systems. Changed one of the solid styles to use a css pattern.


wareid (approved):

gregoriopellegrino (approved):
Tested on my Kobo ereader, work great!

PR #235, opened 2022-10-04
Additional FXL Layout tests

iherman:
Three new tests:

- Viewport w/h are set to device-width and device-height, respectively
- No viewport is set (_should_ use device-width and device-height)
- Erronuous viewport is set (_should_ use device-width and device-height)

Edited on 2022-10-15: as a result of https://github.com/w3c/epub-tests/pull/235#issuecomment-1274449290 the content of the PR has changed:

- Viewport w/h are set to device-width and device-height, respectively
- Reading systems SHOULD attempt to extract 'width' and 'height' values even if the tag's syntax is invalid.
- Reading system should use the first declaration of a 'width' or 'height' (i.e., ignore repeated declarations).

Also, an older test has been updated to use patterns instead of colors only, to make it usable for b/w/ displays.


wareid (approved):

tests/lay-fxl-xhtml-icb_error/EPUB/page_001.xhtml
```diff
@@ -0,0 +1,12 @@
+<html xmlns="http://www.w3.org/1999/xhtml" xmlns:epub="http://www.idpf.org/2007/ops" xml:lang="en">
+  <head>
+    <title>Invalid viewport setting yields device dimensions</title>
+    <meta name="viewport" content="anything, width=1000, height=1000" />
```
davemanhall, line 4: This brings up an interesting question - if we can successfully parse out width and height from this (which we do), should we use it?  I believe the intent of the test is that it should default to device width/height.  Anyhow, a discussion to be had somewhere else.  :-)

davemanhall (approved):
Looks good - able to open all tests.

tests/lay-fxl-xhtml-icb_error/EPUB/page_001.xhtml
iherman: Right, this is an issue (and the test reflects the current spec), and it is being discussed in https://github.com/w3c/epub-specs/issues/2442 about now... We would love to have your reactions, @davemanhall

iherman:
These tests may have to be changed/adapted to the outcome of https://github.com/w3c/epub-specs/pull/2453. Should be considered as a draft for the time being.

tests/lay-fxl-xhtml-icb_invalid_meta/EPUB/page_002.xhtml
```diff
@@ -0,0 +1,12 @@
+<html xmlns="http://www.w3.org/1999/xhtml" xmlns:epub="http://www.idpf.org/2007/ops" xml:lang="en">
+  <head>
+    <title>Invalid Meta — 2</title>
+    <meta name="viewport" content="width=900, height=600, something, foo" />
```
rdeltour, line 4: There are a variety of other ways to produce invalid viewport metadata, including (not exhaustively):
- `"something, width=900, height=600"`
- `"width=900, something, height=600"`
- `"width=900, height=600,,"`
- `"something=width=100,width=900, height=600"`
- `"width=,width=900, height=600"`
- etc…
Should this be tested too?

rdeltour (commented):
As suggested in a comment inline: the test case for invalid viewport meta only tests one possible (invalid) pattern. More cases can be conceived, but I don't know if the intent is to be thorough or to have only one key test.

tests/lay-fxl-xhtml-icb_invalid_meta/EPUB/page_002.xhtml
iherman: To be honest, it sounds a bit over the top to create separate tests for all these. 

I think the emphasis there is to see what the RS does if the viewport parser is indeed incorrect. I would think it is a reasonable assumption that the RS would react the same way with all the variations that you describe there (of course, one can imagine an implementation that concentrates on the particular issue in this very test, but that is not a realistic assumption). The test shows (if RS-s pass it, that is) that the particular normative statement can be implemented, that the specification is implementable by independent implementations, and, in a sense, this is the fundamental goal of the test suite (but, of course, it should also be useful for RS-s in general).

Bottom line: I think we are fine.

PR #239, opened 2022-10-07
Extra report flag: "n/a"

### LINKED ISSUE #237: Differentiate non-applicable tests from untested tests in the report
I think we had mentioned this at TPAC, but I don't remember if we had arrived at a solution.

Right now in the report we can enter three values: `true`, `false` and `null`.

The thing is that `null`, in my case means either not applicable, or not yet tested (or to be rechecked because there are problems with the reader).

Can it make sense to split these two different pieces of information?

iherman:
As discussed in #237, the scripts have been modified to accept a "n/a" entry in the implementation report file, signalling a "non applicable" for that specific test. The value of "null" stands for "not yet tested".

It so happens that the Wysebee test results already used this flag (which made the report not conformant:-) but that means there is a realistic view of how things would look at:

https://raw.githack.com/w3c/epub-tests/Extra-response-flag/docs/drafts/results.html#sec-detailed-publication-resources-results

I was not sure whether the "?" mark is ok as a cell content for not tested; any idea for a better sign is welcome (I tried to fill the cell with the text "not tested", but it made the table too "full".

Fix #237


iherman:
(As a side effect I found a bug in the script which has nothing to do with the new feature; this PR gets that handled, too.)

wareid (approved):

src/lib/data.ts
```diff
@@ -194,8 +194,21 @@ function consolidate_implementation_reports(implementations: ImplementationRepor
             // Minor side effect trick: for each key there is at least one Test Result that has a value. We can therefore
             // safely set the result to 'false' temporarily if the test result is not available; there must be a false or true
             // value somewhere else in the array anyway.
-            const all_results: boolean[] = variant_results.map((results: TestResults): boolean => key in results ? results[key] : false);
-            retval[key] = all_results.reduce( (p: boolean, c: boolean): boolean => p || c, false);
+            const all_results: (boolean|string)[] = variant_results.map((results: TestResults): (boolean|string) => {
```
bduga, line 197: These types are getting a little ugly. If you wanted to you could add a map step that converts the Boolean|string to an Enum with Untested, Pass, Fail, NA values. But entirely optional, fine as is.

bduga (approved):

src/lib/data.ts
iherman: I agree it is something to be done. At this moment, I would add it to a TODO list for after merging this PR; it is important to introduce the functionality into the system and then do this change behind the scenes, so to say.

PR #241, opened 2022-10-07
the two icb tests which failed initially passes if the view port is increased

clapierre:
IE. I don't see the grid until I increase the dimensions of Thorium app itself to be bigger than the default window size it starts in.  However the multi_declarations and units tests didn't have this same issue.


iherman:
I had some similar problems with Thorium, actually, but that seems to be some issues with the grid construction in CSS. Ie, it is not a matter of the test's subject...

PR #242, opened 2022-10-11
Enum usage in the script

iherman:
This PR is a purely coding clean-up issue; reacting to the comment of @bduga (https://github.com/w3c/epub-tests/pull/239#pullrequestreview-1134885040)  the code now uses, internally, a proper `enum` instead of the ugly `boolean|string` combinations.


src/lib/data.ts
```diff
@@ -118,40 +127,58 @@ export async function get_list_dir(dir_name: string, filter_name: (name: string)
...
+    type internal_index_pair = [string, Score];
+    interface raw_map {[index: string]: (boolean|string)}
+    interface internal_map {[index: string]: Score}
+   
+    // the boolean|string in the JSON file is transformed into a proper Score
+    const transform = (raw_score: (boolean|string)): Score => {
+        if (typeof(raw_score) === 'boolean') {
+            return (raw_score) ? Score.PASS : Score.FAIL;
+        } else {
+            return (raw_score === "n/a")? Score.NONAPPLICABLE : Score.UNTESTED;
```
bduga, line 147: Maybe raw_score.toLowerCase() === "n/a" ? Unless that happens somewhere else and I am just missing it.

src/lib/types.ts
```diff
@@ -136,6 +136,39 @@ export interface TestData {
...
 
+
+/**
+ * (Internal) values for the test scores
+ */
+// eslint-disable-next-line no-shadow
+export enum Score {
+    FAIL = "fail",
+    PASS = "pass",
+    NONAPPLICABLE = "n/a",
```
bduga, line 147: Nit: technically N/A means not applicable. Also not sure what the naming conventions are here, but maybe it should be NOT_APPLICABLE?

src/lib/types.ts
```diff
@@ -136,6 +136,39 @@ export interface TestData {
...
+    UNTESTED = "?",
+}
+
+export namespace Score {
+    export function get_class(s: Score): string {
```
bduga, line 152: Nice!

src/lib/types.ts
```diff
@@ -136,6 +136,39 @@ export interface TestData {
     required: ReqType;
 }
 
+
+/**
+ * (Internal) values for the test scores
+ */
+// eslint-disable-next-line no-shadow
```
bduga, line 143: That's worrisome - what is being shadowed?

bduga (commented):
Sorry, I missed this until I saw the latest mention of me in another review :( 

Main question below is about the lint check.

src/lib/data.ts
iherman: No, it is not handled elsewhere and it is indeed a good idea. Added.

src/lib/types.ts
iherman: I have searched all over the place in my code and there is no such 'Score' symbol anywhere. Worse: whatever symbol I use instead, the effects are the same, ie, this seems to be an eslint bug. And that is _really_ worrysome, because it casts a shadow on all eslint warnings...

src/lib/data.ts
iherman: Typescript requires me an extra "else if (raw_score === null)" to do so, though, because technically the toLowerCase() does not work on null. And typescript is absolutely right!

src/lib/types.ts
iherman: Yep, done

PR #243, opened 2022-10-13
Add test for Thorium 2.0 on Windows 10

gautierchomel:
Tests done in june so tests added since are null. Will be updated with Thorium 2.2.


iherman (approved):
Thanks @gautierchomel. 

I would prefer to delay merging this until #242 is merged, because otherwise it will lead to ugly merge conflicts. I will wait with the merge of #242 until the week end, maybe @bduga will have the time to look at that one, so that we an merge both.

PR #250, opened 2022-11-12
Vertical rl test

toshiakikoike:
Since some RSs cannot mix horizontal and vertical writing-mode,I separated XHTML for horizontal and vertical writing-mode or I set whole text vertical writing-mode.


iherman (approved):
I leave the detailed technical review to @wareid, she is more familiar with this. Thank you for updating the `dcterms:modified` fields in the metadata, it is important.

One question: I see you have added the `page-progression-direction` attribute to the `spine` element. It is a good idea, but I wonder whether there is (whether there should) be a difference in the RS behavior by doing this. Put it another way, is it worth repeating at least some of the tests _without_ that attribute test to see if the RS do the right thing.

(That attribute is set to be optional, so it may be o.k. to ignore this question.)

tests/lay-reflow-align-x-center/EPUB/content_001.xhtml
```diff
@@ -1,6 +1,12 @@
 <html xmlns="http://www.w3.org/1999/xhtml" xmlns:epub="http://www.idpf.org/2007/ops" xml:lang="en">
   <head>
     <title>align-x-center in reflow</title>
+    <style>
+      html {
+        -epub-writing-mode:vertical-rl;
+        writing-mode:vertical-rl;
+      }
+    </style>
```
wareid, line 9: Does this test need the text to be vertical? I worry about adding additional complexity when the test doesn't require it.

wareid:
I agree with Ivan's comment as well (added another too). Overall the tests look fine, I just worry about added unnecessary complexity with the changing of page progression and writing-mode for the tests that do not require it. If RS is rendering text differently depending on those I'd consider that worth reviewing since they should be unrelated.

tests/lay-reflow-align-x-center/EPUB/content_001.xhtml
toshiakikoike: "align-x-center" was added to respond to the request in Japanese vertical writing books.
Isn't this function necessary for horizontal writing?

The sample I made during the IDPF era was also written vertically.
https://github.com/toshiakikoike/epub-testsuite/tree/master/content/30/epub30-test-0130

https://github.com/w3c/epub-specs/issues/1380#issuecomment-719108924
As I wrote this issue, I think that only Voyager's BinB passes this test.

wareid: If you want to leave it, that's fine, I just didn't want to overcomplicate the test.

iherman:
@wareid if you are fine with the tests as they are, go ahead and merge it...

wareid:
@toshiakikoike can you please generate the EPUB files for these and replace the ones in the folder currently?

PR #251, opened 2022-11-14
Added two naviagtion document tests, with an image in the navigation instead of text 

iherman:
This is to add test to PR https://github.com/w3c/epub-specs/pull/2479. If this test is o.k., should be merged when that one is merged.


iherman:
Note that a third test is updated (`lay-pkg-flow-scrolled-continuous`); that is only a spelling mistake change, no relationship. I was lazy creating a separate PR only for that one...

tests/nav-non-text_img/EPUB/content_001.xhtml
```diff
@@ -0,0 +1,13 @@
+<html xmlns="http://www.w3.org/1999/xhtml" xmlns:epub="http://www.idpf.org/2007/ops" xml:lang="en">
+  <head>
+    <title>Navigation document with an image and alt text</title>
+  </head>
+  <body>
+    <p>Test passes if the table of content in the reading system contains two links with the following labels:</p>
```
mattgarrish, line 6: It would pass if the image shows, too (for this and the next test).

I don't expect that will be the case often, but we should rule out false positives if someone does test an html-based reading system. The test should probably be something like, if the image shows, mark this test as N/A, otherwise, the text label needs to be displayed.

(And I was wrong about finding a reading system that supported images. When I tried with these tests it failed both showing the image and the fallback text. And when I re-ran my sample, I didn't see the image again. No idea what I thought I was seeing...)

iherman: +1, I have added a sentence marking the rendered image case as n/a

PR #257, opened 2022-12-03
Added test results form ACCESS and VOYAGER

shiestyle:


iherman:
@shiestyle As submitted, these two reports will stay separate in the final report, even in the consolidated results' table. The reason is that the `"name"` fields in the two reports are different, and the consolidation is based on a common name. Is this the intention, or are these two implementation sharing the same core engine, and they have an android and a web variant? If the latter, then you should use a common name for the two reports.

iherman (approved):

shiestyle:
Yes, they are different.

wareid (approved):

PR #259, opened 2022-12-06
Improved the test reports 

iherman:
This is a follow-up of the discussion held on [Dec. 2](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2022-12-02-epub). I have made some changes on the way the implementation results are presented. These are as follows:

- (as agreed on the call) the "consolidated" vs. "detailed" implementation result tables are now separated into two, top level sections. Also, due to the way the table of content is done in the W3C style, the separation is much clearer now.
- Another issue that is not obvious is the fact that there are features (e.g., Media Overlay) that are _not_ required for a conformant RS, which also means that, strictly speaking, these features are not to influence the fate of the Proposed Recommendation. On the other hand, these features have a number of "must" tests in there, which are, sort of "optional" must statements. To handle these more easily when the Director has to review the results, I have done the following changes:
    - I reordered the features and pushed them to the end of the list (these are, in the current setup, Scripting, Media Overlays, and Structural Semantics)
    - I generate an extra, standard text into their respective section, namely:

    > The general feature is _OPTIONAL;_ a "must" tests means that _it is required to pass it to claim conformance in implementing the feature_.
    - (This may be contentious, but it is easy to take it out.) These subsections do _not_ appear when the button is used to switch the visibility of "must" vs. "must/should/may" tests.
- I have marked the rows in the consolidated tables that do not have at least two "pass" cells, ie, that are problematic for CR. (This is done by adding a class name to the row and I have added a CSS rule, i.e., the exact ways of presenting these are easy to change.) We may decide _not_ to leave this in the final report, but it may help us in the meantime.

You can see the result of the changes on a [preview](https://htmlpreview.github.io/?https://github.com/w3c/epub-tests/blob/improve-report/docs/results.html).

cc @wareid @bduga @davemanhall @TzviyaSiegman @johnfoliot @clapierre @avneeshsingh @mattgarrish @dlazin


wareid (approved):

iherman:
Closing this in favour of #263

PR #262, opened 2022-12-07
Tests for Google Play Books

bduga:
Some more reports for Google that I had never submitted.


iherman:
@bduga, could you look at those merge conflicts, please? I cannot make any changes, because the incoming files are from your repo...

(You probably started with an earlier version)

bduga:
Yes. it's on my list :) I thought I had synced, but probably not.

PR #265, opened 2022-12-07
Updated Voyager's result for EPUBCheck errors

shiestyle:
Changed "false" to "n/a" for untested cases due to EPUBCheck errors


wareid:
@shiestyle The `pub-cmt-x` tests are all for content types, did they really fail? Or does Voyager not support embedded media?

wareid:
Oh shoot, let me fix those. We can merge this for now.

PR #266, opened 2022-12-07
Fixing epubcheck errors not related to tests

wareid:
Fixing all of the errors in epubcheck that aren't related to the tests themselves.


dauwhe (approved):
I think a copy of this PR should be sent to:

1. Whoever decided that IDs can't start with numbers
2. Whoever thought up boolean attributes in HTML5

PR #272, opened 2023-01-10
Remove question marks from the tables

iherman:
This PR proposes a tiny change in the test result tables: replace the '?' character with a simple space. The color coding of the cells, as well as the text labels ('pass', 'fail', 'n/a') does not make it necessary to use the '?' character, and an empty cell just does it. On the other hand, I found the large number of '?' characters tiring to the eye...

### DIFF
--- a/docs/drafts/results.html
+++ b/docs/drafts/results.html
@@ -21,8 +21,12 @@
 				background-color: hsl(355, 100%, 72%);
 			}
 
+			td.untested {
+				background-color: hsl(60, 100%, 85%);
+			}
+
 			td.na {
-				background-color: hsl(60, 100%, 79%);;
+				background-color: hsl(30, 5%, 97%);
 			}
 
 			tr.under_implemented {
--- a/src/lib/html.ts
+++ b/src/lib/html.ts
@@ -133,7 +133,7 @@ function create_impl_reports(data: ReportData): {consolidated_results: string, c
                 for (const result of row.implementations) {
                     if (result === undefined) {
                         // This may happen if the tester has not started with a full template...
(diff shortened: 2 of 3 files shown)


mattgarrish:
Why not just put in a label like 'unverified' and give it a background style? Empty holes in the table are equally distracting, especially when the changing row colours come through. It would make the result part of the table more uniform if there were a consistent set of styles and labels to every cell.

(I'd probably use the yellow background for the unverified cells and change n/a to grey, as it's common to have inapplicable stuff diminished, not highlighted.)

iherman:
> Why not just put in a label like 'unverified' and give it a background style? Empty holes in the table are equally distracting, especially when the changing row colours come through. It would make the result part of the table more uniform if there were a consistent set of styles and labels to every cell.
> 

With the number of cells horizontally, the term "unverified" will be too long and will probably be broken into lines. Any term with max four characters?


> (I'd probably use the yellow background for the unverified cells and change n/a to grey, as it's common to have inapplicable stuff diminished, not highlighted.)

I like that, I will do this color change

mattgarrish:
> Any term with max four characters?

To do?

(That's a tough restriction!)

iherman:
> > Any term with max four characters?
> 
> To do?
> 
> (That's a tough restriction!)

:-)

> 
> Edit: "Todo" is also an acceptable word by some sources if you don't want the space.

I will run with that for now.

iherman:
Ok. I have made the changes. There is no preview in the PR, but here is how it looks like now (screen dump on my machine):
Screenshot 2023-01-18 at 17 33 56

PR #278, opened 2023-01-12
Add report for EasyReader covering MO tests

mattgarrish:
I didn't use the test files for verifying support as EasyReader is more particular about how media overlay publications are constructed (all text must be synchronized, etc.), but I was able to mimic all the tests with a separate publication.

Ideally, this report should be more comprehensive, but given the primary need for verifying media overlays I'm starting only with that section. If time permits, I can try and fill in some more general support details.


mattgarrish:
As there isn't a diff or preview, the media overlay test reporting begins at line 59 of the file: https://github.com/w3c/epub-tests/pull/278/files#diff-a55b9edd5fbec56da548b8e9bdd894676312ee08e9ab48764d533215798acf65R59

iherman:
I am a bit worried about submitting a test report for which most cells will be a '?' (or an empty cell, if #272 is accepted). It will look a bit strange...

I would think we should merge this as a last resort. I have just received, separately, a colibrio report that seems to cover most of the MO tests, so the whole issue may be moot... (I will submit the separate PR with that in a moment.)

PR #280, opened 2023-01-12
Removing fxl-orientation-duplication test

iherman:
This is not really a test for the RS, it expliitly refers to epubcheck. This should be tested as part of the structural tests.


iherman:
@mattgarrish is that feature handled by epubcheck? (It should...)

mattgarrish:
Yes, it throws an error on this test file about the duplication.

mattgarrish (approved):

PR #277, opened 2023-01-11
Update Apple Books iOS media overlay support

mattgarrish:
To verify the FXL alternatives I created in https://github.com/w3c/epub-tests/pull/274, I went through loading them in Apple Books on an iPad and have updated the report file with the results.

PR #285, opened 2023-01-13
Added the fxl test equivalents

iherman:
There was an issue raised on the last call regarding the MO related tests, namely that many RS in fact implement MO on fixed layout documents only. Back on the call the MO test results did not look good, because most tests had only one implementation (from Thorium, thanks @gautierchomel for the report)

Since then the situation improved because we have now a Colibrio report as well that covers all the MO tests (thanks @larscwallin) so the MO section is not in jeopardy as far as the W3C process is concerned. Nevertheless, it may be better to show the situation with fixed layout documents; this is a valuable information for the community which we should make clear.

This PR adds 6 new MO tests (created by @mattgarrish) that are duplicates of existing MO tests with the only difference that the documents are pre-paginated rather than reflowable. I have then run these tests through Thorium, Colibrio, and also Apple Books on the Mac (@mattgarrish has already run them on Apple books IOS, see #277), and they all run o.k. As a consequence, this PR also modifies the Colibrio, Thorium, and Apple Books reports by adding entries for the new tests. (The Apple Books report gives a `false` to, say, `mol-support_xhtml` and `true` to `mol-support_xhtml-fxl`).

@bduga, I was not sure how to run these tests through Google Books; I wonder whether you could do that. @wareid, the same question for Kobo. It would certainly look way better if both Google Books and Kobo would run successfully on these.

To ease the job, @bduga and @wareid the 6 new tests are:

- `mol-support_xhtml-fxl`
- `mol-support_xhtml-load-fxl`
- ~`mol-support_xhtml-load-next-fxl`~
- `mol-support_xhtml-load-next-fxl`
- `mol-timing-synchronization_multiple_audio-fxl`
- `mol-timing-synchronization_svg-fxl`


mattgarrish (approved):

bduga:
I will try and run them today.

bduga:
And just to verify my arithmetic - it's just 5 tests, right? It looks like one of them was duplicated in the list.

iherman:
> And just to verify my arithmetic - it's just 5 tests, right? It looks like one of them was duplicated in the list.

Oops, sorry...

davemanhall (approved):
Great work, looks good!

PR #287, opened 2023-01-16
pkg-linked-records should be a "SHOULD"

### LINKED ISSUE #284: Test pkg-linked-records - According to the spec. this test should be a "SHOULD"
"Retrieval and support of [linked resources](https://www.w3.org/TR/epub-33/#dfn-linked-resource) is OPTIONAL."

https://www.w3.org/TR/epub-rs-33/#sec-pkg-doc-metadata

iherman:
Fix #284


iherman:
@larscwallin is it o.k.? (Cannot set you as an official reviewer...)

PR #289, opened 2023-01-18
Wrong compression was used for ocf-zip-comp

iherman:
The test was supposed to use a disallowed compression for the epub file, and the reader is supposed to report an error and not open the file. However, the epub on the repository used the proper compression... Handled now.

As for the test results: Books on Mac OS and Thorium reacted as required, producing an error report. The MacOS test report was already `true`; Thorium gave a `false` and I now changed it to `true`.

Colibrio did not display the book but produced no error message. The spec just says that other compressions should not be used, so I will set the Colibrio results to `true`, too (although an error report would be nice...).


iherman:
cc: @davemanhall @gautierchomel @larscwallin

PR #290, opened 2023-01-18
Expanded on the explanation for the mol-audio-no-clipend test

### LINKED ISSUE #276: mol-audio-no-clipend confusing instructions
This test only says that the audio has to play to the end. It leaves out that highlighting stops with the third sentence.

It would help to say that the third sentence has no clipend so watch that the audio keeps playing with it remaining highlighted (i.e., even though the following text never gets highlighted).

The test failed by not starting audio playback of the third sentence, but I wasn't sure what was going on when the audio stopped after the second sentence.

iherman:
The title tells it all...

Fix #276


wareid (approved):

tests/mol-audio-no-clipend/EPUB/content_001.xhtml
```diff
@@ -11,7 +11,10 @@
         <li>If turned on, the extract on the next page is read aloud by the Reading System</li>
         <li>The full text is read</li>
     </ol>
-    <p>The Reading System may highlight the text being read with a green background with the text grayed out;
-        but this feature is not required for the test to pass.</p>
+    <p>
+        The Reading System may highlight the text being read with a green background with the text grayed out;
+        but this feature is not required for the test to pass. Note that, if highlighted,
+        the highlight may get out of sync with after the second sentence: the audio keeps playing while the highlight remains on the second sentence. This is a possible effect of the missing <code>clipEnd</code> for the second sentence.
```
mattgarrish, line 17: It could be a bit more direct about what will happen as there are only two par elements to synchronize with (it's also the third sentence, as the first and second are merged into a single span):

```suggestion
        the highlighting will stop at the third sentence while the audio keeps playing to the end of the paragraph. This is the effect of missing a <code>clipEnd</code> on the corresponding `par` element in the media overlay file.
```

mattgarrish (approved):

Issue #293, opened 2023-03-02
Test confreq-rs-file-urls -  Reading Systems must prevent access to resources referenced via file URLs


GJFR:
I'd like to add a test for https://www.w3.org/TR/epub-rs-33/#confreq-rs-file-urls:
> Reading Systems MUST prevent access to resources referenced via file URLs

This test will attempt to access resources on the user's local file system. If the reading system allows access, the test fails.

However, there's some complications that have to be addressed, and I'd like some input on their solutions such that they are sure to be in line with the general testing guidelines:

- This test has to take into account OS file system differences (e.g., Windows paths look like `file:///C:/...` while UNIX paths look like `file:///...`). In this regard, we could make a subtest for every OS variant:
  - `pub-file-urls_macos`
  - `pub-file-urls_linux`
  - `pub-file-urls_windows`
  
  Or simply put all in one single test, and differentiate over sections.

- [Our 2021 study](https://gjfr.dev/pdf/epub_ieee_ssp) showed that some reading systems load local `file://` resources depending on resource type (e.g., text, image) or file extension (e.g., .log, .txt, .mp3). However, testing for all resource types and file extensions might be a bit overkill and complicated. In my opinion, testing for access to local text and image files would already give a representative result.

- By leveraging pre-existing _well-known_ OS files that exist on virtually any OS instance (independent of user settings), I think we can prevent this test to require the tester to manually create files on their OS. Examples of these pre-existing files are:
  - MacOS
    - Text: `file:///var/log/system.log`
    - Image: `file:///System/Library/CoreServices/DefaultBackground.jpg`
  - Linux
    - Text: `file:///var/log/system.log`
  - Windows
    - Text: `file:///C:/Windows/setuplog.txt`

Let me know if there's any concern about these suggestions.

iherman:
> I'd like to add a test for https://www.w3.org/TR/epub-rs-33/#confreq-rs-file-urls:
> 
> > Reading Systems MUST prevent access to resources referenced via file URLs
> 
> This test will attempt to access resources on the user's local file system. If the reading system allows access, the test fails.
> 
> However, there's some complications that have to be addressed, and I'd like some input on their solutions such that they are sure to be in line with the general testing guidelines:
> 
> * This test has to take into account OS file system differences (e.g., Windows paths look like `file:///C:/...` while UNIX paths look like `file:///...`). In this regard, we could make a subtest for every OS variant:
>   
>   * `pub-file-urls_macos`
[...]

PR #294, opened 2023-03-03
Add pub-file-urls test

GJFR:
Add test for checking whether reading systems prevent access to file URL resources (#293, [spec](https://www.w3.org/TR/epub-rs-33/#confreq-rs-file-urls)).

Sanity check done for readings systems on MacOS (e.g. calibre, Apple Books, colibrio, EPUBReader). I also checked some outdated reading systems that are known to load file URL resources, where the test functioned correctly as well.

I will have access to Windows and Linux this evening or tomorrow, so I can test on these operating systems as well.


GJFR:
Great, @iherman! Will prepend the requested URL for each iframe.

I'd be happy to update the test reports. However, as I currently don't have an Android smartphone at hand, I might not be able to test all reading systems.

Just a small question; next to the `true` and `false` test outcomes, there's also `null`. When should the latter be used?

iherman:
> Great, @iherman! Will prepend the requested URL for each iframe.
> 
> I'd be happy to update the test reports. However, as I currently don't have an Android smartphone at hand, I might not be able to test all reading systems.
> 
> Just a small question; next to the `true` and `false` test outcomes, there's also `null`. When should the latter be used?

There are actually two extra possible outcomes beyond the obvious true/false:

- null : not yet tested
- "n/a" : non applicable for some reasons (e.g., reading systems that does not have side-loading would use that for your side-loading tests)

If you cannot test all the reading systems, do not worry about it. The original authors should actually do that. What is really important at this point that, hopefully, each of the new tests will have at least two positive test results (the more the merrier of course), because otherwise we would have to do something about the corresponding MUST statement if it is a MUST statement we are testing (if not, it is not really an issue).

iherman:
Iforgot to say: the default is null, ie, if the test reference doesn't appear in the list. So if you cannot test, don't worry about adding anything to the respective file.

iherman (approved):
The final approval is contingent on having some implementation reports also filled in. At that point, I will be happy to merge.

GJFR:
It's difficult to find reliable images that are by default available on any OS variant. I guess it's best to leave it out for now, and just check for access to the log files. Later on, the test could extended to look for other files (perhaps requiring the tester to create files on pre-defined paths).

I will be able to fill in the reports tonight or tomorrow, and complete this PR.

Issue #295, opened 2023-03-03
Test - Reading systems that allow local storage SHOULD provide methods for users to inspect or delete that data.


GJFR:
I will write a test for:

> Reading systems that allow [local storage](https://html.spec.whatwg.org/multipage/#dom-localstorage) [[html](https://www.w3.org/TR/epub-rs-33/#bib-html)] SHOULD provide methods for users to inspect or delete that data.

(https://www.w3.org/TR/epub-rs-33/#sec-local-storage)

PR #296, opened 2023-03-03
Add scr-storage-delete test

GJFR:
Add test for checking whether users can delete persistent data in local storage set by an EPUB (#295, [spec](https://www.w3.org/TR/epub-rs-33/#sec-local-storage)).

- Tester should open the EPUB two times, such that the script can indicate whether persistent data is supported (some reading systems support local storage, but wipe it upon termination of the reading session).
- If persistent storage is supported, tester should look for reading system user interface options to delete this data.

I think this is the most straightforward way to test this?


iherman (commented):
My (very minor) comment is on the key chosen for the persistent storage (`random_data_entry`) which bears the (minor) danger of colliding with other EPUB's using persistent storage for some reason. Maybe something more test specific, like `scr_storage_delete_random_data_entry` is better? Just to be on the safe side...

The final approval is contingent on having some implementation reports also filled in. At that point, I will be happy to merge.

GJFR:
Completely agree!

GJFR:
This one should be ready 👍

Issue #299, opened 2023-03-07
Test - every EPUB publication should be assigned a unique origin


GJFR:
I will add a test for:
> It MUST assign a unique origin [html], shared by all spine-level scripts of the EPUB publication.

(https://www.w3.org/TR/epub-rs-33/#confreq-rs-scripted-origin-shared)

This test requires opening an EPUB publication (which will show its assigned origin), closing it, and subsequently opening another EPUB publication (which will also show its assigned origin). The test will pass if both origins are different.

If I'm correct, there seems to be no procedure for creating a test that relies on two EPUB publications. As such, I think it's best to re-use the EPUB publication of the existing `scr-support_origin` test as one of the two EPUB publications that will be opened.

PR #300, opened 2023-03-07
Add unique origin test

GJFR:
Add test for checking whether the reading system assigns a unique origin to each EPUB publication (#299, [spec](https://www.w3.org/TR/epub-rs-33/#confreq-rs-scripted-origin-shared)).


iherman (commented):
I love the idea of re-using another bona fide test! 😀

Happy to merge once the implementation reports are there.

iherman:
@GJFR this looks increasingly good :-).

Please, push the "ready to review" button when you consider this done...

GJFR:
I wanted to run more tests but ran into some issues:
- Some reading systems only execute inline scripts within an iframe, some only external scripts, etc. So to make this test (and by extension all other tests that rely on scripting) as compatible as possible, a script should be included in multiple ways.
- Some reading systems use an `opaque` or `null` origin. This is great from a security perspective, since it means that scripts and persistent data are isolated, but I'm not sure if this is desirable from a usability perspective (i.e. not sure if localStorage among the chapters of an EPUB is shared when origin is `opaque` or `null`).
- Because web browsers do not allow a website to change its origin, web readers will not be able to comply with this requirement of having a unique origin for each EPUB publication. Or is there something I'm missing?

I think it's best if these are discussed in separate issues because they might affect other tests?

iherman:
> I wanted to run more tests but ran into some issues:
> 
> * Some reading systems only execute inline scripts within an iframe, some only external scripts, etc. So to make this test (and by extension all other tests that rely on scripting) as compatible as possible, a script should be included in multiple ways.

Let us not forget that the whole scripting feature set is _optional_. So we can add additional tests, but I would postpone that for later (my worry is to go to a PR stage with half finished things and that will be the target to problems...)

> * Some reading systems use an `opaque` or `null` origin. This is great from a security perspective, since it means that scripts and persistent data are isolated, but I'm not sure if this is desirable from a usability perspective (i.e. not sure if localStorage among the chapters of an EPUB is shared when origin is `opaque` or `null`).

From the spec perspective, what counts is the [root url](https://www.w3.org/TR/epub-33/#sec-container-iri) combined with [the roog URL in the reading system](https://www.w3.org/TR/epub-rs-33/#sec-container-iri). If that is correct then we are fine...

But I readily acknowledge that I am slightly (perhaps not slightly) out of my depth here, and I would like @rdeltour to look at this. He is the real expert (most of the text in those sections come from him, actually...) 

> * Because web browsers do not allow a website to change its origin, web readers will not be able to comply with this requirement of having a unique origin for each EPUB publication. Or is there something I'm missing?

I guess this is a typical case where the test result should say "n/a". Web based RSs are simply a different thing altogether in this respect.

> 
> I think it's best if these are discussed in separate issues because they might affect other tests?

GJFR:
Okay, thanks for the clarifications! I've updated the tolino-webreader report to `n/a`.

This should be ready to be merged.

PR #303, opened 2025-04-15
Update to 3.4

iherman:
The test suite has been adapted to EPUB 3.4; more generally, to be able to move the test suite to a higher version. The following changes are important:

- The requirements for the tests' package document have changed a bit. It
    - includes a reference to "version", meaning a version of the spec in development at the time the test is created
    - the references to the specs (from the package document) refer to the specs' short name
-  The generating scripts take care of using the right URLs in the test tables (referring to the 3.4 editors' drafts), and a separate column has been added to the test table with the version (that can be used to reorder the table rows). The hope is that this will make the CR phase and reporting easier, showing the "inherited" tests as well as the new ones.
    
    The package documents of all current tests have been updated.
- In line with the changes of the specs, the reports also use dark mode

I have also made some housekeeping on the underlying scripts (updated the npm packages to the latest versions, reviewed and slightly improved/modernized the scripts itself, updated the contributors' documentations, etc.)


mattgarrish (approved):
This is one of those too massive to read PRs!

I say let's go with out usual approach to these kinds of changes and merge now and work out any kinks later as we go.

shiestyle (approved):

PR #304, opened 2025-04-18
Generation update

iherman:
This PR makes a major change in the underlying repository structure, but no change whatsoever on the tests or testing methodology. Two changes

- I exchanged the node.js environment to [deno](https://deno.land). Deno provides a more comfortable TypeScript environment and, mostly, a stronger and stricter TS interpreter (which, rightfully, found some brittle details in my code while converting). The code itself has only changed superficially.
- The previous setup had a disagreeable feature: the gh action script generated intermediate files, used for the report, which then were pushed back to the main repo. The net result was that all cloned repositories were always out of sync and had to be pulled without no good visible reasons. The new version makes use of a new facility for GitHub pages, which generates the github.io content directly from the action script without touching the repository itself. This needed some reorganization of the repository (details are boring) but the result is that the aforementioned annoyance is gone.


iherman:
This is one of the "too big to review" PR-s; merging it and see the possible problems later...

PR #306, opened 2025-05-26
Extended the cnt-css-fonts test to include legacy media type

iherman:
This to adapt to https://github.com/w3c/epub-specs/pull/2726.


mattgarrish:
I think this might be trying to get too specific. You're testing CSS support for TTF font rendering but suggesting that the package document media type influences the rendering. Is that ever (reliably) the case?

I don't know if there are any meaningful reading system support tests for the media-type attribute, but maybe @bduga or @danielweck have thoughts on this? I can only think of content validation tests.

bduga:
> I don't know if there are any meaningful reading system support tests for the media-type attribute, but maybe @bduga or @danielweck have thoughts on this? I can only think of content validation tests.

The only thing I can think of is package level fallbacks, but those would probably fail anyway and are discouraged in this specific case in favor of CSS fallbacks.

Though I am not necessarily against this change. A RS that used the media type as the source of truth for font types would properly render this content. I don't think a RS like that exists in reality, but it is at least possible.

iherman:
> A RS that used the media type as the source of truth for font types would properly render this content.

I.e., from the testing point of view, this test is fine, right?

I realize that this test addresses two slightly different things: (1) accepting font files with the combination of the right category (woff, OT, etc) and media type and (2) correct handling of the CSS technique to really _use_ those fonts. But I have difficulties to see how one could dissociate the two aspects, hence my choice of extending the test (which, originally, concentrated only on the CSS features combines with various categories of fonts).

One could actually argue that testing CSS is out of scope for us anyway; the test suite does not intend to duplicate web platform tests, only on the specifically EPUB usage of the web platform...

mattgarrish:
> But I have difficulties to see how one could dissociate the two aspects

Because you're assuming the reading system is going to block a resource based on its media type declaration in the package document. That's not what core media types are for. They define resources you don't have to provide a fallback for. Nowhere does the spec say that reading systems should block resources based on their media type attribute declaration. The closest requirement is to swap in a manifest fallback resource, as Brady says, but that's never worked for CSS and the reading system has to defer to CSS fallback rules in this case (i.e., it shouldn't block the resource even if it doesn't think it can be handled because the CSS processor will take care of what to display).

So if the test fails, it's failing because CSS is not supported, because font imports aren't supported, or because the font type is not supported, not because of the media type declaration. All the test appears to do differently now is check multiple times that core media type fonts are supported in CSS.

mattgarrish:
On a different note, shouldn't this be split up into separate tests? I realize it's testing the requirement to support core font media types, but wouldn't it be better to determine which one(s) are failing than have a blanket test for all of them?

iherman:
> > But I have difficulties to see how one could dissociate the two aspects
> 
> Because you're assuming the reading system is going to block a resource based on its media type declaration in the package document. That's not what core media types are for. They define resources you don't have to provide a fallback for. Nowhere does the spec say that reading systems should block resources based on their media type attribute declaration. The closest requirement is to swap in a manifest fallback resource, as Brady says, but that's never worked for CSS and the reading system has to defer to CSS fallback rules in this case (i.e., it shouldn't block the resource even if it doesn't think it can be handled because the CSS processor will take care of what to display).
> 
> So if the test fails, it's failing because CSS is not supported, because font imports aren't supported, or because the font type is not supported, not because of the media type declaration. All the test appears to do differently now is check multiple times that core media type fonts are supported in CSS.

Hm. In other words, what you are saying is that the test is unnecessary... Which is fine with me, we can drop it and stay with the old one, which was merely testing whether woff, woff2, ot, or tt were all working (using CSS). However, how would we test the normative statements on media types then? (I realize that fonts are not the only media types that do not have tests...)

iherman:
>> Hm. In other words, what you are saying is that the test is unnecessary...
>
> Not the whole test. I was just concerned this was trying to capture something we can't verify.

Sorry, I was not clear. The original test had the same structure, just using a single media type per font format. Ie, it had one line each for woff, woff2, OT, and TT. What is unnecessary is the expansion to format-media type combinations.

> But we still have a requirement that reading systems must support the font formats from font-face rules, so the tests are still useful there. 

Which is what the original test did. We were/are covered.

> We only need one test for each required font format, though.

Yeah, that was the weakness of the original test. 

I think that:

- this PR should be closed without merging
- a new PR should be created that takes the original test and splits it into 4 similar tests, one font format per test. (And the test reference in the RS has to be changed accordingly.)

Correct? 

As a penitence, I will do the test split at some point...

mattgarrish:
> Correct?

Sounds right to me.

danielweck:
> Is it normal that `application/font-sfnt` appears twice in this table?

Answering my own question: https://github.com/w3c/epubcheck/issues/339

PR #313, opened 2026-01-31
PP and roll tests with svg in spine

iherman:
I reproduced the pp and roll layout tests of the ABC book, using SVG in the spine instead of XHTML. (Using SVG seems to be simpler than with XHTML when simply embedding images.)


iherman (commented):
At this moment (2026-01-31) it seems to work well on Apple Books. It does not seem to work out of the box with Thorium (version v3.3.0 on MacOS): the consecutive pages seem to be rendered on top of one another without spread, ie, it displays ever second page.

PR #314, opened 2026-06-25
Create four new tests for mixed-layout publication test cases

reidmore-online:
Created four new tests to have more complete coverage of mixed layout EPUB use cases. 

* `lay-pp-spine-overrides_image-only-pp` - A pre-paginated publication with image-only fixed layout pages, where one spine item is reflowable 
* `lay-pp-spine-overrides_image-only-reflow` - A reflowable publication with image-only fixed layout pages, where one spine item is pre-paginated 
* `lay-pp-spine-overrides_image-spine-pp` - A pre-paginated publication with image-in-spine fixed layout pages, where one spine item is reflowable
* `lay-pp-spine-overrides_image-spine-reflow` - A reflowable publication with image-in-spine fixed layout pages, where one spine item is pre-paginated


HadrienGardeur:
Thank you @reidmore-online, this will be helpful.

On Slack you also mentioned two other test files ([fxl-spine-overrides_behave-as-global](https://github.com/w3c/epub-tests/tree/main/tests/fxl-spine-overrides_behave-as-global) and [fxl-spine-overrides_behave-as-global-bis](https://github.com/w3c/epub-tests/tree/main/tests/fxl-spine-overrides_behave-as-global-bis)) that were included in the EPUB 3.3 process, do we plan on adding these two tests to our EPUB 3.4 test coverage as well?

reidmore-online:
Those tests are also part of the 3.4 test suite, they are just renamed to fit in with the new structure.

HadrienGardeur:
All good then, if we also keep the previous tests we should be well-covered thanks to these additions.

HadrienGardeur (approved):

tests/lay-pp-spine-overrides_image-only-pp/EPUB/page_001.xhtml
```diff
@@ -0,0 +1,10 @@
+<html xmlns="http://www.w3.org/1999/xhtml" xmlns:epub="http://www.idpf.org/2007/ops" xml:lang="en">
+  <head>
+    <title>Override with a reflowable spine item: pre-paginated page</title>
+    <meta name="viewport" content="width=1000, height=600" />
+    <link rel="stylesheet" href="fixed.css" />
+  </head>
+  <body>
+  	<img src="images/page_1.png" alt="This test passes if this page is displayed as fixed layout, and the next page is displayed as reflowable layout." />
```
iherman, line 8: Shouldn't there be a text somewhere describing when the test passes in a pure text? The alt text may not be visible, and it makes it unclear when it passes and when not...

tests/lay-pp-spine-overrides_image-spine-pp/EPUB/page_001.xhtml
```diff
@@ -0,0 +1,10 @@
+<html xmlns="http://www.w3.org/1999/xhtml" xmlns:epub="http://www.idpf.org/2007/ops" xml:lang="en">
+  <head>
+    <title>Override with a reflowable spine item: pre-paginated page</title>
+    <meta name="viewport" content="width=1000, height=600" />
+    <link rel="stylesheet" href="fixed.css" />
+  </head>
+  <body>
+  	<img src="images/page_1.png" alt="This test passes if this page is displayed as fixed layout, and the next page is displayed as reflowable layout." />
```
iherman, line 8: Same comment as above...

iherman:
See https://github.com/w3c/epub-specs/pull/3031. That PR should be merged after this one.

tests/lay-pp-spine-overrides_image-only-pp/EPUB/page_001.xhtml
reidmore-online: The text is in the image itself, and it needs to be in the image as this is an image-only test and adding text would result in different behaviour for several reading systems.

tests/lay-pp-spine-overrides_image-spine-pp/EPUB/page_001.xhtml
reidmore-online: Might be the same case for above or not, but the text is in the image itself.

sueneu:
I don't have an understanding of the testing process, I'm

tests/lay-pp-spine-overrides_image-only-pp/EPUB/page_001.xhtml
iherman: Ah, sorry, I did not realize the image is a text itself. All good, my mistake.

iherman (approved):

tests/lay-pp-spine-overrides_image-only-reflow/EPUB/page_001.xhtml
```diff
@@ -0,0 +1,156 @@
+<html xmlns="http://www.w3.org/1999/xhtml"
+    xmlns:epub="http://www.idpf.org/2007/ops" xml:lang="en">
+    <head>
+        <title>Override with a pre-paginated spine item: reflowable document</title>
+    </head>
+    <body>
+        <p>Test passes if this and the third page displays properly as a reflowable page, whereas the second page is displayed as a pre-paginated page.</p>
```
mattgarrish: I wonder about using "page" here to describe the behaviour. It makes sense if the reflowable pages are scrolled, but if they're dynamically paginated then it's not the first and third pages that reflow but the first and third spine items. Maybe it should be more generic, like if there is a pre-paginated document with an image between two reflowable text documents then it passes?

(The other tests where the pre-paginated document comes after a reflowable are likewise going to potentially be misleading about what to look for.)

sueneu: I figure "page" is a legacy term borrowed from print, and generally understood in context of the structure of a "book". Since ebooks pick up much of their structure from printed books, "page" doesn't seem out of place even if it isn't an exact match for the technology. 

The generic phrasing plus using the itemrefs would get around the page/screen/veiw question. 
><p>Test passes if the pre-paginated content of page_002 is displayed between the reflowable text in page_001 and page_002.</p>

or if order is helpful
><p>Test passes if the book is displayed with the content of page_001 as reflowable, then the content of page_002 as pre-paginated, followed by the content of page_003 as reflowable.</p>

reidmore-online: Let me see if I can make this a little clearer.

reidmore-online:
@mattgarrish let me know if this change addresses the question better!

PR #317, opened 2026-07-24
Adding an 'tested-by' flag to the test report.

iherman:
As discussed on our [telco](https://w3c.github.io/pm-wg/minutes/2026-07-23.html#61ef): there is now an optional extra boolean flag in the report, called `official`. Its effect is on the [list of implementations](https://w3c.github.io/epub-tests/results#sec-implementer-list) in the report. If the flag is present, and its value is `true`, and extra remark is added. See the screenshot of my local test run:

Screenshot 2026-07-24 at 08 54 58

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -243,7 +243,7 @@ JSON file is as follows:
 
 * `ref` (optional): A URL that creates a link on the name of the reading system in the implementation report.
 
-* `official` (optional): If set to `true`, the tests have been performed by a persons that officially represent the implementers of the reading system for the purpose of testing.
+* `tested-by` (optional): If set to `implementer` the tests have been performed by the implementers of the reading system, otherwise it should be set to `third-party`.
 
 * `tests`: An object with the list of the implementation results. Each key is a test's unique identifier (its `dc:identifier`) with a
   value of `true`, `false`, `"n/a"`, or `null` for a test that passes, fails, is not applicable (i.e., is not implemented for some reasons), or is not yet tested, respectively. If a test is not listed, its value is considered to be `null`. The implementation report will show "?" for `null`, indicating that the implementation has not run the test.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -255,7 +255,7 @@ Here is an example of a small test report:
     "name"     : "ACME Books",
     "ref"      : "https://www.example.org/acme",
     "variant"  : "iOS, v1.0",
-    "official" : true,
+    "tested-by" : "implementer",
     "tests" : {
         "pub-cmt-gif": true,
         "pub-cmt-jpeg": true,
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -243,16 +243,19 @@ JSON file is as follows:
 
 * `ref` (optional): A URL that creates a link on the name of the reading system in the implementation report.
 
+* `official` (optional): If set to `true`, the tests have been performed by a persons that officially represent the implementers of the reading system for the purpose of testing.
+
 * `tests`: An object with the list of the implementation results. Each key is a test's unique identifier (its `dc:identifier`) with a
-  value of `true`, `false`, `"n/a"`, or `null` for a test that passes, fails, is not applicable (i.e., is not implemented for some reasons), or is not yet tested, respectively. If a test is not listed, or its value is considered to be `null`. The implementation report will show a value of "?" for `null`, indicating that the implementation has not run the test.
+  value of `true`, `false`, `"n/a"`, or `null` for a test that passes, fails, is not applicable (i.e., is not implemented for some reasons), or is not yet tested, respectively. If a test is not listed, its value is considered to be `null`. The implementation report will show "?" for `null`, indicating that the implementation has not run the test.
 
 Here is an example of a small test report:
 
 ```json
 {
-    "name"  : "ACME Books",
-    "ref"   : "https://www.example.org/acme",
-    "variant" : "iOS, v1.0",
+    "name"     : "ACME Books",
+    "ref"      : "https://www.example.org/acme",
+    "variant"  : "iOS, v1.0",
+    "official" : true,
     "tests" : {
         "pub-cmt-gif": true,
         "pub-cmt-jpeg": true,
--- a/generate/src/deno.jsonc
+++ b/generate/src/deno.jsonc
@@ -1,7 +1,7 @@
 {
     "description": "Generation of index and resolutions' files for EPUB tests and test results.",
-    // Version 1.1.0 adds the concept and management of deprecated tests
-    "version": "1.1.1.",
+    // Version 1.2.0 adds the extra "official" flag to the test results
+    "version": "1.2.0.",
     "compilerOptions": {
         "lib": ["dom", "dom.iterable", "dom.asynciterable", "esnext", "deno.ns"]
     },
--- a/generate/src/deno.jsonc
+++ b/generate/src/deno.jsonc
@@ -1,6 +1,6 @@
 {
     "description": "Generation of index and resolutions' files for EPUB tests and test results.",
-    // Version 1.2.0 adds the extra "official" flag to the test results
+    // Version 1.2.0 adds the extra "tested-by" field to the test results
     "version": "1.2.0.",
     "compilerOptions": {
         "lib": ["dom", "dom.iterable", "dom.asynciterable", "esnext", "deno.ns"]
--- a/generate/src/lib/data.ts
+++ b/generate/src/lib/data.ts
@@ -174,10 +174,11 @@ async function getAnImplementationReport(fname: string): Promise<ImplementationR
     const raw_data = await Deno.readTextFile(fname);
     const raw_report: Raw_ImplementationReport = JSON.parse(raw_data) as Raw_ImplementationReport;
     return {
-        name    : raw_report.name,
-        ref     : raw_report.ref,
-        variant : raw_report.variant,
-        tests   : transform_tests(raw_report.tests),
+        name     : raw_report.name,
+        ref      : raw_report.ref,
+        variant  : raw_report.variant,
+        official : raw_report.official,
+        tests    : transform_tests(raw_report.tests),
     }
 }
 
@@ -543,8 +544,9 @@ export function getTemplate(report: ReportData): Raw_ImplementationReport {
     }
 
     return {
-        "name"  : "(Implementation's name)",
-        "ref"   : "https://www.example.com",
-        "tests" : test_list,
+        "name"     : "(Implementation's name)",
+        "ref"      : "https://www.example.com",
+        "official" : false,
+        "tests"    : test_list,
     }
 }
--- a/generate/src/lib/data.ts
+++ b/generate/src/lib/data.ts
@@ -20,6 +20,7 @@ import {
     Implementer, ReportData,
     ReqType,
     Constants,
+    TESTED_BY,
 } from './types.ts';
 
 
@@ -174,11 +175,11 @@ async function getAnImplementationReport(fname: string): Promise<ImplementationR
     const raw_data = await Deno.readTextFile(fname);
     const raw_report: Raw_ImplementationReport = JSON.parse(raw_data) as Raw_ImplementationReport;
     return {
-        name     : raw_report.name,
-        ref      : raw_report.ref,
-        variant  : raw_report.variant,
-        official : raw_report.official,
-        tests    : transform_tests(raw_report.tests),
+        name        : raw_report.name,
+        ref         : raw_report.ref,
+        variant     : raw_report.variant,
+        "tested-by" : raw_report["tested-by"] || TESTED_BY.IMPLEMENTER,
+        tests       : transform_tests(raw_report.tests),
     }
 }
 
@@ -544,9 +545,9 @@ export function getTemplate(report: ReportData): Raw_ImplementationReport {
     }
 
     return {
-        "name"     : "(Implementation's name)",
-        "ref"      : "https://www.example.com",
-        "official" : false,
-        "tests"    : test_list,
+        "name"      : "(Implementation's name)",
+        "ref"       : "https://www.example.com",
+        "tested-by" : TESTED_BY.IMPLEMENTER,
+        "tests"     : test_list,
     }
 }
--- a/generate/src/lib/html.ts
+++ b/generate/src/lib/html.ts
@@ -6,7 +6,7 @@
  */
 
 import type { ReportData, Implementer, HTMLFragments } from './types.ts';
-import { Constants, Score, ReqType }                   from './types.ts';
+import { Constants, Score, ReqType, TESTED_BY }        from './types.ts';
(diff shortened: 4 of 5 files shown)


sueneu (approved):

reidmore-online (approved):

shiestyle (approved):

generate/src/lib/html.ts
```diff
@@ -61,7 +61,9 @@ function createImplementationList(impl: Implementer[]): string {
     for (const implementer of impl) {
         const li = addChild(ol, 'li');
         const name = 'ref' in implementer ? `<a href="${implementer.ref}">${implementer.name}</a>` : `${implementer.name}`;
-        li.innerHTML = 'variant' in implementer ? `${name}, ${implementer.variant}` : name;
+        const fullName = "variant" in implementer ? `${name}, ${implementer.variant}` : name;
+        const official = "official" in implementer && implementer.official === true ? " (official test results)" : "";
```
bduga: Nit: Maybe drop the `in` check for the existential operator (or whatever it is called in TS)

```suggestion
        const official = implementer?.official === true ? " (official test results)" : "";
```

bduga (approved):

bduga:
And having thought about this a little more, maybe the term `official` could be seen as scary to some contributors? If I am a random engineer at a big company, providing an "official" response might be problem. Maybe `from_implementer` or `from_developer` or something like that?

sueneu:
>Maybe from_implementer or from_developer or something like that?

I see where you're going here. Implementer = reading system  developer = Ebook software developer/maker ?

bduga:
> I see where you're going here. Implementer = reading system  developer = Ebook software developer/maker ?

Yeah, isn't that what we are trying to encapsulate with this property? That this came from internal testing. I just worry that the term official may scare away some internal testers.

sueneu:
>I just worry that the term official may scare away some internal testers.

Or make other testers feel less valued

iherman:
Can we agree in the change " (official test results)" -> " (test results by implementers)" ?

mattgarrish:
> " (official test results)" -> " (test results by implementers)"

Would it be simpler to label the "unofficial" reports instead? A "(third-party test results)" label on submissions that didn't come from anyone directly involved with the reading system might be easier to understand.

iherman:
> > " (official test results)" -> " (test results by implementers)"
> 
> Would it be simpler to label the "unofficial" reports instead? A "(third-party test results)" label on submissions that didn't come from anyone directly involved with the reading system might be easier to understand.

I am fine with that approach, too, if it is ok with the majority... I can go either way.

bduga:
I am also fine with either approach, though I think we still want to change the `official` property and description.

reidmore-online:
I like the idea of marking only those not provided directly by implementers, as in practice we've typically mostly gotten tests from them first. 

Possible option: `tested-by` with values `implementer` or `third-party`?

iherman:
I actually like what @reidmore-online's proposal in https://github.com/w3c/epub-tests/pull/317#issuecomment-5104657356:

> `tested-by` (optional): Can be `implementer` or `third-party`

I think I will go with that. Default is "implementer" and the extra remark will be added if the value is "third-party". Actually, code-wise, the net effect is that if the value is ***not*** "implementer", then it will end up with a remark... because at the moment there is no error reporting on the submission and the default mechanism means that "implementer" will win if the value is anything but that. But I do not think this is a major problem in this case...

~~Minor difference, though: Typescript does not like object keys to include a "-" sign. It is possible, but awkward. The usual alternative is "tested_by", which is what I will use.~~ (I decided to keep to `tester-by`; the code is a bit uglier but that is all right.)

iherman:
It now uses the proposal of https://github.com/w3c/epub-tests/pull/317#issuecomment-5104657356 by @reidmore-online. This is how it looks now:

Screenshot 2026-07-29 at 10 01 58

@reidmore-online I have not changed the README.md file, I do not want to create possible merge conflicts with #318. We can do that later.

iherman:
Please re-review @bduga @sueneu @mattgarrish @reidmore-online

sueneu:
Ivan's example looks good to me (edited for more specificity)

generate/src/lib/types.ts
```diff
@@ -269,7 +285,14 @@ export interface Implementer {
     variant ?: string;
 
     /** If present, the name becomes a hyperlink to this URL. */
-    ref     ?: string
+    ref     ?: string;
+
+    /**
+     * If present and set to true, the tester is "official", ie,
```
bduga, line 291: This comment needs to be updated.

generate/src/deno.jsonc
```diff
@@ -1,7 +1,7 @@
 {
     "description": "Generation of index and resolutions' files for EPUB tests and test results.",
-    // Version 1.1.0 adds the concept and management of deprecated tests
-    "version": "1.1.1.",
+    // Version 1.2.0 adds the extra "official" flag to the test results
```
bduga:
```suggestion
    // Version 1.2.0 adds the extra "tested-by" field to the test results
```

CONTRIBUTING.md
```diff
@@ -243,16 +243,19 @@ JSON file is as follows:
 
 * `ref` (optional): A URL that creates a link on the name of the reading system in the implementation report.
 
+* `official` (optional): If set to `true`, the tests have been performed by a persons that officially represent the implementers of the reading system for the purpose of testing.
```
bduga:
```suggestion
* `tested-by` (optional): If set to `implementer` the tests have been performed by the implementers of the reading system, otherwise it should be set to `third-party`.
```

Leaving this vague, since there can be a lot of people we might consider implementers (e.g. significant contributor to an open source project).

CONTRIBUTING.md
```diff
@@ -243,16 +243,19 @@ JSON file is as follows:
...
 
 ```json
 {
-    "name"  : "ACME Books",
-    "ref"   : "https://www.example.org/acme",
-    "variant" : "iOS, v1.0",
+    "name"     : "ACME Books",
+    "ref"      : "https://www.example.org/acme",
+    "variant"  : "iOS, v1.0",
+    "official" : true,
```
bduga:
```suggestion
    "tested-by" : "implementer",
```

generate/src/lib/data.ts
```diff
@@ -174,10 +175,11 @@ async function getAnImplementationReport(fname: string): Promise<ImplementationR
...
     const raw_report: Raw_ImplementationReport = JSON.parse(raw_data) as Raw_ImplementationReport;
     return {
-        name    : raw_report.name,
-        ref     : raw_report.ref,
-        variant : raw_report.variant,
-        tests   : transform_tests(raw_report.tests),
+        name        : raw_report.name,
+        ref         : raw_report.ref,
+        variant     : raw_report.variant,
+        "tested-by" : raw_report["tested-by"] || TESTED_BY.IMPLEMENTER,
```
bduga, line 181: You might want to avoid hyphens here, which will be both consistent with variable names elsewhere and remove the need for quotes here. Same change in other places for consistency.

```suggestion
        "tested_by" : raw_report.tested_by || TESTED_BY.IMPLEMENTER,
```

bduga, line 181: Is this the correct default to use? It seems to me we should only mark this as performed by an implementer if they actually claim to be one, but maybe you have it this way to deal with the more common case? I guess it is up to you if you think this will be typical, but it does mean that anyone who failed to provide the property is now an implementer.

generate/src/lib/types.ts
```diff
@@ -256,6 +256,22 @@ export namespace Score {
...
+export enum TESTED_BY {
+    IMPLEMENTER = "implementer",
+    THIRD_PARTY = "third-party",
+}
+
+export namespace TESTED_BY {
+    export function get_tester(t: TESTED_BY | undefined): string {
+        // This is a bit loose; any string that is not "implementer" will lead to a third party flag.
+        // But this looseness is not really harmful for our purposes...
+        return t !== undefined && `${t}` !== "implementer" ? "(tested by a third party)" : "";
```
bduga, line 272: I may just be TypeScript naive, but do you need to do the string conversion? And perhaps use the type value instead of a raw string?

```suggestion
        return t !== undefined && t !== TESTED_BY.IMPLEMENTER  ? "(tested by a third party)" : "";
```

And if this field isn't optional (see comment at fied definition) you can drop the undefined check, which is kind of confusing.

generate/src/lib/types.ts
```diff
@@ -269,7 +285,14 @@ export interface Implementer {
...
     /** If present, the name becomes a hyperlink to this URL. */
-    ref     ?: string
+    ref     ?: string;
+
+    /**
+     * If present and set to true, the tester is "official", ie,
+     * the test have been performed by someone who officially represent
+     * the implementers
+     */
+    "tested-by" ?: TESTED_BY;
```
bduga, line 295: Ok I have to confess I haven't read through all the code, but I think this optionality is a red herring that is making your code overly complex. From what I have seen, when we read the raw data and convert it to this interface, we always provide this field. You could remove the `?` and then all your checks for undefined are unnecessary.

bduga (commented):
I am good with the general idea, though see comments for some questions about the default value for implementer. Plus some other little things.

CONTRIBUTING.md
reidmore-online: Would this be `"tested_by" : "implementer",` if your other changes are accepted?

bduga: It would. My suggestions (here and elsewhere) all use the current code in case there is a reason not to take take the other suggestion (hyphen to underscore).

generate/src/lib/types.ts
iherman: Unfortunately, this did not work (it was my first setup). I must admit I did not understand why, but I know that the enumerations in Typescript has some strange behavior. The usage of `${t}` is a bit ugly, but it has the advantage that it works...

generate/src/lib/data.ts
iherman: Yes, that is how I started. The problem is that asking the end user to use `tested_by` may sound unnatural while, for example, `-` is used, for example, for test names. Hence my opting for `-` in the test input. But that JSON is directly parsed into its internal representation, so the natural think is that the object key would be "tested-by".

The important thing is to make the end user (ie, the tester) happy, even if this means a slightly uglier code.

iherman: If I remember it right, most (all?) the 3.3 test results were submitted by the implementers. So yes, this seems to be the common case (and also our preferred case) hence this default.

@reidmore-online, do I remember well for the 3.3 testers?

generate/src/lib/types.ts
iherman: Yeah... I started with this enumeration, I thought it would be helpful (as it is for other cases) and yes, at the end of the day, it may not be so important or helpful. I may find some time to make the whole things work in terms of plain strings; but this version has the advantage of already working... :-)

CONTRIBUTING.md
iherman: But I may prefer keeping this with '-'. See https://github.com/w3c/epub-tests/pull/317#discussion_r3676037820.

I will accept this change because it is a bug. We can update later.

generate/src/lib/data.ts
bduga: Yes, I see your older comment now (I had missed it previously), and I understand there seems to be some sort of dislike for underscore in the "we type raw JSON" world, but that doesn't mean we have to use the same value in our code. So tested-by in the JSON can become tested_by in the code, which both pleases the typists and cleans up the code. Actually, I think internally this should really be testedBy, which seems to be the standard TS naming convention.

generate/src/lib/types.ts
bduga: That just plain worries me, and implies something else is wrong. I just tested this and it seems to work, see this horrific URL: https://www.typescriptlang.org/play/?#code/FAAgpgdgrgtiAqBRAykgIgfQEIE0QG9QRiBJAWQAUAZRMxAOSQCUQBeEAIgEsYAHAGzAxIAFzAAnDgBoixeAAkSTTBQCCTeHnYcRACy7iAJgFpeAQ3EiAntOABfYMABmUCAGMRXAPYQQAczARDDEAZzFxAAoRAC4EFHRsPAAfEFdDMCcuCDBDAEpYsPEsvwJZEAB6coR9EJAuWrMQACMuERB+Ly8QsABuEDMIKxBC4pA9Mzb6kAgvNu4+QWEIcI4QAHcufn52sDNDMa9+sf0jEHNLIad+Mz8AOjLKkCwoNr0pjq7IMBDaqZm28S7LZDXQWGAubZOLziEBeKAw3jw3ifEK3NFlQEieG+NoAQlY7DSGSyORAADIyWMQPj2EhUIhMLhbuRqLQGMwQAB+ThRb5ifZNIaNN6nc7WXKrWIcDg9eyOABuFiptPiDMStwUShU6k0wDcPhCXkEtw6fgiASCoXCUVyuWAbRV9MZOGZlBodEYiCYeoNRrAJq8ZotwT5EhtdqAA

generate/src/lib/types.ts
bduga: Sorry, I wasn't clear. I am fine with the enumeration as-is (and prefer it to raw strings), my comment was about the optionality of the field. Specifically, why not:

```suggestion
    "tested-by" : TESTED_BY;
```
That would make any further checks for undefined unnecessary. For instance, the signature for get_tester could be simplified to:
```
function get_tester(t: TESTED_BY): string
```

As far as I have seen, even if the field is missing in the raw data, it is always present in the internal structure.

generate/src/lib/data.ts
iherman: Well we would gain something and loose something in the code. At the moment, I take the result of JSON parse as is, ie, to produce an instance of the object type defined for the code. Changing this would mean a virtual duplication+transformation of the raw input and the internal type definitions. All this extra complication for... not much. That is why I prefer the current approach.

generate/src/lib/types.ts
iherman: It is a bit like the answer above: at the moment, at least for the "metadata" aspect, the raw and the internal data share the TS datatypes. This would mean separating them. So gaining some complexity vs loosing some...

generate/src/lib/types.ts
iherman: @bduga I will look into this. However, I do not think I can do it in the coming days. If it is all right with you, I have raised a separate issue (https://github.com/w3c/epub-tests/issues/319) and I would prefer to merge this PR rather than waiting on this. Who knows, maybe we will get a test submitted soon, and I would like to be ready for it...

generate/src/lib/data.ts
bduga: I don't think it is really any additional work, since you are already doing the transformation manually, but also it is fine to leave it. It isn't wrong, just a bit wonky.

generate/src/lib/types.ts
bduga: Ok. This one worries because it clearly *should* work and the fact that it doesn't means there is another problem in the code we don't understand. I may poke around here a little bit over the next day or two.

PR #318, opened 2026-07-28
Adding more detailed testing instructions

reidmore-online:
Updating the README and repo to have an issue template for submitting results. 

Depends on the decision about PR #317, tentatively I have `tested-by` but will change depending on what we decide. 

I can also adjust the instructions section to match in the `contributing` page.


iherman:
I have a setup problem. At the moment, we have an entry point for [contributions](https://w3c.github.io/epub-tests/contributing), which is linked from the general index.html page. That file is, actually, respec wraparound for [contributing.md ](https://github.com/w3c/epub-tests/blob/main/CONTRIBUTING.md) in the repository. Shouldn't that the place where the changes should be made? (Maybe increasing the visibility of that page in the readme is necessary...)

iherman:
The problem I have with the template is that, afaik, if there is a template then there is no way to "just" create an issue. It all have to go through the template. That means if someone wants to add a new test, comment on one, or simply raise an issue/PR on, say, the test reporting tool, it becomes a problem.

I may be wrong, but it is worth checking.

README.md
```diff
@@ -15,3 +15,88 @@ The [test reports](https://w3c.github.io/epub-tests/) describe all the tests, in
...
+- `variant` (optional): The name or properties of the reading system variant. Typical values may be Android, iOS, or Web, if one implementation (i.e., sharing the same name value) has specific versions running in those environments. In addition to the properties of the reading system, testers may also include information on how the tests were run, for example, if a screen reader was used with the reading system.
+- `ref` (optional): A URL that creates a link on the name of the reading system in the implementation report.
+- `tested-by` (optional): Can be `implementer` or `third-party`, this field is to help track whether the test results were provided by the platform or a third-party tester. 
+- `tests`: An object with the list of the implementation results. Each key is a test's unique identifier (its dc:identifier) with a value of true, false, "n/a", or null for a test that passes, fails, is not applicable (i.e., is not implemented for some reasons), or is not yet tested, respectively. If a test is not listed, or its value is considered to be null. The implementation report will show a value of "?" for null, indicating that the implementation has not run the test.
```
iherman:
```suggestion
- `tests`: An object with the list of the implementation results. Each key is a test's unique identifier (its `dc:identifier`) with a value of true, false, "n/a", or null for a test that passes, fails, is not applicable (i.e., is not implemented for some reasons), or is not yet tested, respectively. If a test is not listed, or its value is considered to be null. The implementation report will show a value of "?" for null, indicating that the implementation has not run the test.
```

iherman (commented):
See my other comments before merge.

README.md
```diff
@@ -15,3 +15,88 @@ The [test reports](https://w3c.github.io/epub-tests/) describe all the tests, in
...
+Make a copy of the template file: [xx-template.json](https://w3c.github.io/epub-tests/reports/xx-template.json), change the file name to reflect the reading system and platform being tested (if multiple are being tested). 
+
+`acme-ios.json`
+
+Open the file in any text editor application and fill in the reporting fields: 
+
+- `name`: The name of the reading system.
+- `variant` (optional): The name or properties of the reading system variant. Typical values may be Android, iOS, or Web, if one implementation (i.e., sharing the same name value) has specific versions running in those environments. In addition to the properties of the reading system, testers may also include information on how the tests were run, for example, if a screen reader was used with the reading system.
+- `ref` (optional): A URL that creates a link on the name of the reading system in the implementation report.
+- `tested-by` (optional): Can be `implementer` or `third-party`, this field is to help track whether the test results were provided by the platform or a third-party tester.
```
iherman:
```suggestion
- `tested-by` (optional): Can be `"implementer"` or `"third-party"`, this field is to help track whether the test results were provided by the platform or a third-party tester. 
```

As you remark somewhere else, the `"` character is significant...

README.md
sueneu: I wonder if comparing two ranges of text with a final "respectively" could be misunderstood, even though it is perfectly correct. 

>`tests`: An object with the list of the test results. Each key is a test's unique identifier (its `dc:identifier`) with a value of `true`, `false`, `n/a`, or `null.` Use `true` for a test that passes, `false` for a test that fails, `n/a` for a test is not applicable. A test would have a result of `null` if is not implemented or is not yet performed. If a test is not listed its value is considered to be `null`. The implementation report will show a value of "?" for `null`, indicating that the implementation has not run the test.

reidmore-online:
@iherman I mirrored the changes to the Contributing file, with edits as appropriate. Biggest change is that I have flipped the contributing content to put "running tests" first, as that is the content that is most important right now, not the creation of tests.

iherman:
> @iherman I mirrored the changes to the Contributing file, with edits as appropriate. Biggest change is that I have flipped the contributing content to put "running tests" first, as that is the content that is most important right now, not the creation of tests.

i just wonder (but we can deal with this once we have the text finalized) whether it is o.k. to have this duplication. Do you think it is o.k. to import readme.md into the respec file instead of contributing.md, making the latter obsolete?

reidmore-online:
> > @iherman I mirrored the changes to the Contributing file, with edits as appropriate. Biggest change is that I have flipped the contributing content to put "running tests" first, as that is the content that is most important right now, not the creation of tests.
> 
> i just wonder (but we can deal with this once we have the text finalized) whether it is o.k. to have this duplication. Do you think it is o.k. to import readme.md into the respec file instead of contributing.md, making the latter obsolete?

I don't think so, since Contributing also has the information on creating the tests, and README has the information on the repo itself. I only updated the Readme to catch anyone who ends up on the repo page and not the github.io pages, especially while we're requesting testing from people. We can remove it from the README later when we're not actively testing if we want.

iherman:
@reidmore-online do you have any experience with https://github.com/w3c/epub-tests/pull/318#issuecomment-5112627112? (I do not...)

reidmore-online:
> @reidmore-online do you have any experience with [#318 (comment)](https://github.com/w3c/epub-tests/pull/318#issuecomment-5112627112)? (I do not...)

I have a feeling you're right, which is why I've now added a bug/feature template in addition to the results one.

iherman:
> > @reidmore-online do you have any experience with [#318 (comment)](https://github.com/w3c/epub-tests/pull/318#issuecomment-5112627112)? (I do not...)
> 
> I have a feeling you're right, which is why I've now added a bug/feature template in addition to the results one.

But then we should also have a template for a new test submission and/or a change in an existing test... :-(

reidmore-online (commented):
> > > @reidmore-online do you have any experience with [#318 (comment)](https://github.com/w3c/epub-tests/pull/318#issuecomment-5112627112)? (I do not...)
> > 
> > I have a feeling you're right, which is why I've now added a bug/feature template in addition to the results one.
> 
> But then we should also have a template for a new test submission and/or a change in an existing test... :-(  

Done!

iherman (approved):

